PR #12, opened 2023-10-08
Initial formalization of the behaviour for LLM

RudolfMan:
Hi there!
First of all, thank you for this initiative!

I'm working in a project where we want to experiment with various different LLMs, and, while having most of them implemented inside library would be nice, I see that the LLM implementation is actually decoupled from the "LangChain framework", if we formalized the behaviour or a protocol it would enable users of the library to more confidently attach their own LLM implementations and maybe contribute back to the library!

I tried to provide with minimal changes that keeps all of the implementation same, just moves the `@spec` of `ChatOpenAI.call/4` into a behaviour callback.

### DIFF
--- a/lib/llm.ex
+++ b/lib/llm.ex
@@ -1,0 +1,21 @@
+defmodule LangChain.LLM do
+  @moduledoc """
+  A behaviour for LLM
+  """
+
+  alias LangChain.Message
+  alias LangChain.MessageDelta
+
+  @type call_response :: {:ok, Message.t() | [Message.t()]} | {:error, String.t()}
+
+  @callback call(
+              struct(),
+              String.t() | [Message.t()],
+              [LangChain.Function.t()],
+              nil | (Message.t() | MessageDelta.t() -> any())
+            ) :: call_response()
+
+  def call(%model{} = config, messages, functions, callback_fn) do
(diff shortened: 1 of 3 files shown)


lib/llm.ex
```diff
@@ -0,0 +1,21 @@
...
+
+  @callback call(
+              struct(),
+              String.t() | [Message.t()],
+              [LangChain.Function.t()],
+              nil | (Message.t() | MessageDelta.t() -> any())
+            ) :: call_response()
+
+  def call(%model{} = config, messages, functions, callback_fn) do
+    model.call(config, messages, functions, callback_fn)
```
RudolfMan, line 19: This actually looks like a poor man's protocol 😅 

May be it would fit better? and `LangChain.ChatModels.ChatOpenAI` would implement it as something like:
```elixir
defmodule LangChain.ChatModels.ChatOpenAI do
  use Ecto.Schema
  # ...
  embedded_schema do
    # ...
  end
  
  def new(..) do
    ...
  end
  
  # ...
  
  defimpl LLM do
    def call(openai, prompt, functions \\ [], callback_fn \\ nil)
    # ...
  end
end
```
?

brainlid:
I'm going to wait on defining any formal interfaces until there are multiple working implementation examples.

PR #13, opened 2023-10-08
Add dialyxir and specify field types of structs

RudolfMan:
Dialyzer was able to find some errors in specs and even suggest a bugfix 🙂


lib/chains/data_extraction_chain.ex
```diff
@@ -88,7 +89,7 @@ defmodule LangChain.Chains.DataExtractionChain do
       |> LLMChain.add_messages(messages)
       |> LLMChain.run()
       |> case do
-        {:ok, _updated_chain, %Message{role: :function_call, arguments: %{"info" => info}}}
+        {:ok, _updated_chain, %Message{role: :function, arguments: %{"info" => info}}}
```
RudolfMan, line 92: after specifying possible values for message `:role` dialyzer caught this 🙌🏻 
https://github.com/brainlid/langchain/blob/95217c19ea81515b119355f25be66e6b5c2f3bb6/lib/message.ex#L45-L47

RudolfMan:
Dialyzer also suggests that this clause will never happen, but I hesitated to propose to delete as is for now
https://github.com/brainlid/langchain/blob/95217c19ea81515b119355f25be66e6b5c2f3bb6/lib/chains/llm_chain.ex#L400-L403

lib/chains/data_extraction_chain.ex
RudolfMan: now it's fixed by https://github.com/brainlid/langchain/pull/16
but still I think the project would benefit from Dialyzer 🙂

brainlid:
Thank you for the work and interest in helping the project. At this time, I'm not focusing on expanded Dialyzer support. I don't want to leave you or the PR hanging in limbo, so for now, I'm going to close it. Thank you though for your interest!

Issue #11, opened 2023-10-06
Add support for OpenAI-Organization header


MrYawe:
https://platform.openai.com/docs/api-reference/organization-optional

brainlid:
Thanks for the PR! Merged. 

❤️💛💙💜

PR #15, opened 2023-10-10
Adds support for setting the `OpenAI-Organization` header in requests

amokan:
This addresses #11 

Many ways to handle this, but using [Req.Request.put_header/3](https://hexdocs.pm/req/Req.Request.html#put_header/3) seemed to be the most sane approach.

Having said that, I can sense you are doing a bunch of development on the project @brainlid - so no offense if you need to reject it if it does not align with where you are going :)

If there is anything I can do to clean this up, please let me know.


brainlid (approved):
I like it!

I like the "maybe_add_..." idea. Works well. Thanks!

brainlid:
❤️💛💙💜

Issue #14, opened 2023-10-09
I get `{:error, "Unexpected response. {:ok, %LangChain.Chains.LLMChain{ ... }}}` when using the DataExtractionChain


sylvester-roos:
Given the following code in Livebook:

```elixir
itinerary_1_day = """
Day 1:

Arrive in Delft and check-in at the Hotel De Emauspoort, a cozy boutique hotel located in the heart of the city.
"""

itinerary_day_schema_parameters = %{
  type: "object",
  properties: %{
    destination_name: %{type: "string"},
    destination_type: %{type: "string"}
  },
  required: []
}

# Model setup
{:ok, chat} = LangChain.ChatModels.ChatOpenAI.new(%{model: "gpt-3.5-turbo", temperature: 0, stream: false, verbose: true})

# run the chain on the text information
data_prompt = itinerary_1_day

{:ok, result} = LangChain.Chains.DataExtractionChain.run(chat, itinerary_day_schema_parameters, data_prompt)
```

I keep getting the following error when running the cell:

```
[...]

PR #16, opened 2023-10-10
Set pattern match in `DataExtractionChain` to look for `role: :assistant` rather than `role: :function_call`

amokan:
... This appears to be the only valid result at this stage.

I see no place where you would expect to receive a response with `role: :function_call` in the current implementation after calling `LLMChain.run()`.

Context here is #14 as I am able to reproduce the same error with the example in that issue as well as the example provided in the moduledoc for `LangChain.Chains.DataExtractionChain`.


brainlid (commented):
Thanks for the fix!

PR #22, opened 2023-10-20
Basic html loader with crawly

warnero:
- Added Document and DocumentLoader Behaviours
- Added Crawly DocumentLoader


warnero:
Hey @brainlid I wanted to split up my work into smaller chunks so I can get it in (and others can play with the blocks/revamp/etc.). How does this one look?

matthusby:
@brainlid I see this has been sitting for a while.  I am planning on doing some data loading from api's soon, and was wondering if there are plans to integrate this PR or some sort of document in general?

brainlid:
I think this effort has stalled out. I’m open to new work in this area.
What do you need?

matthusby:
I am not doing anything too fancy, just planning to pull in some jira tickets and maybe github issues.

My main question is what do you think of using the Document model that is in this PR? I would like to stick to a standard way of doing the document loading etc, at first glance this seems fine - but wanted to make sure I wasn't missing something.

brainlid:
> what do you think of using the Document model that is in this PR

I think the Document model was incomplete. The idea was to base it on the TS/Python LangChain Document idea. I'm not using it personally nor do I have any short-term needs for it. However, I'm open to that approach.

brainlid:
I'm closing this for now as I think it's not going to go further. Thanks!

Issue #5, opened 2023-09-29
Upgrade Req library


brainlid:
A new version of the Req library was released before this library was published.

Upgrade to the latest Req. v0.4.x.

The API for streaming responses changed.

brainlid:
Completed by PR https://github.com/brainlid/langchain/pull/25

PR #25, opened 2023-10-25
Update req to 0.4.4

medoror:
The tests on main seem to be broken at the moment. I did validate that running `mix test --include live_call` before and after the req upgrade resulted in the no change to test output


Cardosaum:
@medoror, Maybe editing your PR message to also add `Closes #5`? (This would enable [automatically closing the related issue](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword)).

---

One thing that I noticed, we don't have CI for this project yet, would it be desirable to add a simple one to ensure there's no regression on PRs, @brainlid ?

brainlid:
@Cardosaum I went ahead and added the basic CI test runs. Note, it does not, and will not, run live tests against external services.

brainlid (approved):
Thanks for testing this out and doing the updates! 

❤️💛💙💜

PR #30, opened 2023-11-01
Allow functions to update context

C-Sinclair:
## Changes

This PR adds the ability for `Function`'s to update the `custom_context` of the chain by returning a tuple of `{response, context}` 

This will unlock the ability for state to be passed between functions without the need to wrap the chain itself in a GenServer or some similar state management mechanism. 

## Use case

You have functions which augment your Chain's Context with additional information, for instance user metadata. Later in the Chain you then want to use this information to add more context for the LLM to use.

### Example

Define one function which fetches the user's birthday

```elixir
Function.new!(%{
    name: "get_user_birthday",
    description: "Returns the birthday of the user.",
    function: fn _arguments, context ->
      birthday = "22/12/1970"
      # The answer is returned to the LLM for immediate usage, but we also store the date in our context for easy access later
      {birthday, Map.put(context, :birthday, birthday)}
    end
  })
```

Define a second function which fetches other users with the same birthday

```elixir
Function.new!(%{
    name: "get_users_with_birthday",
    description: "Returns users who were born on a given day.",
    function: fn _arguments, context ->
      get_user_born_on(context.birthday)
    end
  })
```

### DIFF
--- a/lib/chains/llm_chain.ex
+++ b/lib/chains/llm_chain.ex
@@ -400,13 +400,24 @@ defmodule LangChain.Chains.LLMChain do
           if chain.verbose, do: IO.inspect(function.name, label: "EXECUTING FUNCTION")
 
           # execute the function
-          result = Function.execute(function, message.arguments, use_context)
+          {result, context} =
+            case Function.execute(function, message.arguments, use_context) do
+              {result, new_context} ->
+                {result, new_context}
+
+              result ->
+                {result, use_context}
+            end
+
+          chain = %LLMChain{chain | custom_context: context}
+
           if chain.verbose, do: IO.inspect(result, label: "FUNCTION RESULT")
 
           # add the :function response to the chain
           function_result = Message.new_function!(function.name, result)
           # fire the callback as this is newly generated message
           fire_callback(chain, function_result)
+
           LLMChain.add_message(chain, function_result)
 
         nil ->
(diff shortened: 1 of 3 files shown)


brainlid:
That's an interesting approach. I'm concerned that putting data into the context like this is doing an end-run around the LLM. I'm not saying "no", but I'd like to understand a good reason for needing this.

In your example, the data should be going through the LLM and not as a separate side channel through the context. For instance, the `get_users_with_birthday` function should take an argument of the date where the LLM decides what day it wants to use. Perhaps during the conversation, the LLM decides it wants the users with the birthday the day before the user.

Context that's relevant for the user that the LLM needs can be embedded in a prompt so the LLM doesn't have to request it. For an example of this, see the [Agent demo of the personal fitness trainer](https://github.com/brainlid/langchain_demo#personal-fitness-ai-agent).

C-Sinclair:
> That's an interesting approach. I'm concerned that putting data into the context like this is doing an end-run around the LLM. I'm not saying "no", but I'd like to understand a good reason for needing this.
> 
> In your example, the data should be going through the LLM and not as a separate side channel through the context. For instance, the `get_users_with_birthday` function should take an argument of the date where the LLM decides what day it wants to use. Perhaps during the conversation, the LLM decides it wants the users with the birthday the day before the user.
> 
> Context that's relevant for the user that the LLM needs can be embedded in a prompt so the LLM doesn't have to request it. For an example of this, see the [Agent demo of the personal fitness trainer](https://github.com/brainlid/langchain_demo#personal-fitness-ai-agent).

That's a good point that the LLM could/should handle the context via the conversation messages themselves 🤔  I'll close the PR as I don't have a solid use case which couldn't be achieved via your suggestion. Thanks!

PR #34, opened 2023-11-08
Adds support for JSON mode in OpenAI Chat model

chrisgreg:
The new update to OpenAI ChatGPT APIs allow users to specify if they want their response to be in JSON or not (defaults to text). 

This allows Langchain to specify the response type as valid JSON or not.

https://platform.openai.com/docs/api-reference/chat/create#chat-create-response_format


brainlid:
This looks good! Thanks for the doc link and the submission.

❤️💛💙💜

Issue #42, opened 2023-12-01
%Mint.TransportError{reason: :closed}


chrisgreg:
I'm intermittently getting %Mint.TransportError{reason: :closed} as a result from `do_api_request/4` - has this been seen before or are there any known ideas why?

I can fire off the exact same request afterwards and it works most of them time but 1% of calls are failing... 

EDIT: My team found that this is an ongoing error with Finch. Look at the latest comments [here](https://github.com/sneako/finch/issues/62).

Are we able to swap the HTTP client out for another?

brainlid:
Should be resolved by PR #43. Let me know if you're still getting the error. For whatever reason, I haven't noticed this as an issue myself.

chrisgreg:
Still getting an error even with the retry - we're using this in production so there's a fair bit of usage coming but this error is still around even with a retry policy wrapping the LLMChain.run call. 

Unsure of the appetite but might be worth swapping to HTTPoison for the underlying caller?

brainlid:
I'm open to it. Thankfully it's a deeper dependency that application code doesn't really see or touch.

brainlid:
I've started getting repeated errors of this.

brainlid:
@chrisgreg I merged an explicit retry experiment into the main branch. If it works for you too, then it could be improved upon.

https://github.com/brainlid/langchain/pull/66/files#diff-98396b383d660bb5274db90a5fd13b1526d5d6b120a2d8ece0e8baf06e55b85eR336

PR #43, opened 2023-12-01
Add retry strategy to OpenAI Chat API requests

chrisgreg:
[Due to the intermittent issues exposed by Finch disconnecting](https://github.com/brainlid/langchain/issues/42), this PR adds Req retry logic to API requests. 

Retries are attempted up to 3 times and trigger for all HTTP errors with a 1 second delay.

I figure a test wasn't required given the Req library tests this and I've just added in options to trigger it.


brainlid (approved):
That seems like a safe approach, especially if it's addressing the issue for you.

PR #50, opened 2023-12-11
Update Req to version 0.46

chrisgreg:
Req version 0.46 retries errors on closed (https://github.com/wojtekmach/req/issues/244).

This should fix the intermittent `%Mint.TransportError{reason: :closed}` error.


chrisgreg:
Actually according to the maintainer - this is now part of 0.4.4 so redeploying my app should do the trick.

brainlid:
Looking at it, it appears you'll have to update your deps to pull from the main branch as it hasn't been published in a release yet.

PR #59, opened 2023-12-18
Add support for Google AI / Gemini Pro model

jadengis:
## Summary

This PR adds a `ChatGoogleAI` model that wraps interactions with the [Google AI Rest APIs](https://ai.google.dev/tutorials/rest_quickstart#multi-turn_conversations_chat) for the purposes of integrating with langchain, thus closing #6.

This change supports the full set of Gemini Pro features, including non-streamed responses, streamed responses and function calling.

### Differences with OpenAI

- Google uses the `"model"` role instead of `"assistant`"
- Streaming chunks have an identical schema to single-shot messages. This allowed a lot of reuse of the message parsing code for both Messages and MessageDeltas.
- I implemented the `for_api/` behaviour using plain old pattern matching. The protocol appoach seems a little roundabout. 

### Quirks

- Google AI doesn't support system prompts at the moment. I asked in the Google Discord server and was told this is a known issue and big ask, but the best thing to do in the meantime is simulate a system message with a user message. This would be great, but the API validates that all user messages in a multi-turn conversation are followed by model message. The way around this is to expand system messages into a user message and shim model message. This is non-ideal and hopefully system prompt support is added soon.
- Streaming text chunks in the SSE format requires the `alt=sse` query param added to the url. This is undocumented, but i noticed it being used in the [official SDKs](https://github.com/google/generative-ai-js).
- Google users `finishReason: "STOP"` for basically everything, including message deltas. This doesn't jive well with some of the existing logic for tracking when e.g. streaming deltas completes. This behaviour is faked in the `ChatGoogleAI` module.

### DIFF
--- a/lib/chains/llm_chain.ex
+++ b/lib/chains/llm_chain.ex
@@ -288,21 +288,35 @@ defmodule LangChain.Chains.LLMChain do
 
     # if the merged delta is now complete, updates as a message.
     if merged.status in [:complete, :length] do
-      case MessageDelta.to_message(merged) do
-        {:ok, %Message{} = message} ->
-          fire_callback(chain, message)
-          add_message(%LLMChain{chain | delta: nil}, message)
-
-        {:error, reason} ->
-          # should not have failed, but it did. Log the error and return
-          # the chain unmodified.
-          Logger.warning("Error applying delta message. Reason: #{inspect(reason)}")
-          chain
-      end
+      delta_to_message(%LLMChain{chain | delta: merged})
     else
       # the delta message is not yet complete. Update the delta with the merged
       # result.
       %LLMChain{chain | delta: merged}
+    end
+  end
+
+  @doc """
+  Convert any hanging delta of the chain to a message and append to the chain.
+
+  If the delta is `nil`, the chain is returned unmodified.
+  """
+  @spec delta_to_message(t()) :: t()
+  def delta_to_message(%LLMChain{delta: nil} = chain) do
+    chain
+  end
+
+  def delta_to_message(%LLMChain{delta: delta} = chain) do
+    case MessageDelta.to_message(delta) do
+      {:ok, %Message{} = message} ->
+        fire_callback(chain, message)
+        add_message(%LLMChain{chain | delta: nil}, message)
+
+      {:error, reason} ->
+        # should not have failed, but it did. Log the error and return
+        # the chain unmodified.
+        Logger.warning("Error applying delta message. Reason: #{inspect(reason)}")
+        chain
     end
   end
 
@@ -314,6 +328,7 @@ defmodule LangChain.Chains.LLMChain do
     deltas
     |> List.flatten()
     |> Enum.reduce(chain, fn d, acc -> apply_delta(acc, d) end)
+    |> delta_to_message()
   end
 
   @doc """
--- a/lib/chat_models/chat_google_ai.ex
+++ b/lib/chat_models/chat_google_ai.ex
@@ -296,12 +296,12 @@ defmodule LangChain.ChatModels.ChatGoogleAI do
     Req.new(
       url: build_url(google_ai),
       json: for_api(google_ai, messages, functions),
-      receive_timeout: google_ai.receive_timeout,
-      finch_request:
-        Utils.finch_stream_fn(google_ai, &do_process_response(&1, MessageDelta), callback_fn)
+      receive_timeout: google_ai.receive_timeout
     )
     |> Req.Request.put_header("accept-encoding", "utf-8")
-    |> Req.post()
+    |> Req.post(
+      into: Utils.handle_stream_fn(google_ai, &do_process_response(&1, MessageDelta), callback_fn)
+    )
     |> case do
       {:ok, %Req.Response{body: data}} ->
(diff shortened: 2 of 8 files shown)


lib/chains/llm_chain.ex
```diff
@@ -307,6 +321,7 @@ defmodule LangChain.Chains.LLMChain do
     deltas
     |> List.flatten()
     |> Enum.reduce(chain, fn d, acc -> apply_delta(acc, d) end)
+    |> delta_to_message()
```
jadengis, line 324: This is fixing a bug I noticed where if the stream only returns a single `MessageDelta` (this happens for streamed function calls for GoogleAI), the tracking delta is never actually converted to a `Message` and the chain will enter an infinite loop.

lib/utils.ex
```diff
@@ -68,4 +70,139 @@ defmodule LangChain.Utils do
     end)
     |> Enum.join("; ")
   end
+
+  @type callback_data ::
```
jadengis, line 74: Pretty much all of this logic was common with the OpenAI code, so I factored it out.

arbaaz:
Nice!!

brainlid:
@jadengis from what I've seen in the code so far, it appears that the server only returns a single version of the assistant's message. For instance, the [OpenAI API has the `n` parameter](https://platform.openai.com/docs/api-reference/chat/create#chat-create-n) for the number of output versions the server should generate. Think like running the same step in the conversation with different seeds and seeing how differently it generates multiple versions.

Anyway, I've never seen another model do that and it complicates the return types. From what you've seen of the Google API, does it have that capability?

I'm considering removing support for that and cleaning up the return type for ChatOpenAI.call

```elixir
  @type call_response :: {:ok, Message.t() | [Message.t()]} | {:error, String.t()}
```

It would just be `{:ok, Message.t()}` instead of an optional array of messages. 

What are your thoughts?

jadengis:
@brainlid I think Google AI API actually does support returning multiple versions of the message. The response JSON for the generation method API contained in[ the docs](https://ai.google.dev/tutorials/rest_quickstart#gemini_and_content_based_apis) contains a `candidates` array which I believe contains should contain all the versions that would be generated. By default it seems to generate only  1 version.

The option for setting this seems undocumented however. It doesn't appear in the [model parameters documation](https://ai.google.dev/docs/concepts#model_parameters), but I did find a `candidatesCount` option in the [official JavaScript SDK](https://github.com/google/generative-ai-js/blob/main/packages/main/types/requests.ts#L60), so I think it should work. That is, sending a request like
```json
{
  "contents": [
    "parts": [{ "text: "User message"}]
  ] ,
  "generationConfig": {
    "candidatesCount": 2
  }
}
```
should return 2 candidates in the response.

I personally don't have a use case for returning multiple versions of a message, but in the interest of keeping things flexible, and since there are two big LLMs that support it, it probably makes sense to leave the type as is. I can think of use cases where being able to generate multiple candidates would be useful

jadengis:
@brainlid Hey is there anyway I can help to get this PR into main? Willing to pitch in if there is any preliminary work required. :pray:

medoror:
@brainlid interested on your thoughts here! I have been looking into integrating [ollama chat](https://github.com/jmorganca/ollama/blob/main/docs/api.md#generate-a-chat-completion) and if this PR is merged, it creates a nice seem for me to start.  

Do you have any hesitations on implementation?

brainlid:
@jadengis I'm not currently setup for testing/verifying the Google endpoint. However, if you're able to help support/fix issues with the integration, then I'm okay to merge it in.

jadengis:
@brainlid  I'm using the Google engine currently in an application, so I've got no problem supporting / fixing issues with the integration. I'll more likely than not need those fixes anyway. As written, it's been working in production without issue for 3 - 4 weeks.

There are a few merge conflicts it looks like. Are there any big changes I should pay attention to in resolving these conflicts? :pray:

brainlid:
@jadengis Sounds good! The merge conflicts should be pretty clean. Can you merge them into your branch?

jadengis:
@brainlid I've updated the PR to be inline with the current `main`.  Updated the Google model to use the same `into` Req trick that you added to OpenAI for streaming. Commoned out the chunk processing code. That's pretty much all the changes. Let me know if it looks good to you :pray:

brainlid (approved):

brainlid:
@jadengis Thanks for all the work you've put into this!

❤️💛💙💜

lib/chat_models/chat_google_ai.ex
```diff
@@ -0,0 +1,432 @@
...
+
+      {:error, changeset} ->
+        {:error, Utils.changeset_error_to_string(changeset)}
+    end
+  end
+
+  def do_process_response(
+        %{
+          "finishReason" => finish,
+          "content" => %{"parts" => parts, "role" => role},
```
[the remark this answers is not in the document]
jadengis: parts contains the actual text segments returned by the LLM. It's normally only a single part actually, but i think if you are using their multi-modal model i.e. `gemini-pro-vision`, it can return both text and image parts in the response. I haven't tried any of the image stuff though. My guess is that the API response format is just future proofed so they can add additional capabilities without needing to break their contracts.

Issue #28, opened 2023-10-29
Support Azure OpenAI


tw4452852:
I've managed to use [Azure OpenAI](https://learn.microsoft.com/en-us/azure/ai-services/openai/reference) with the following minor change:

```diff
diff --git a/lib/chat_models/chat_open_ai.ex b/lib/chat_models/chat_open_ai.ex
index 5dd610e..06b4d0c 100644
--- a/lib/chat_models/chat_open_ai.ex
+++ b/lib/chat_models/chat_open_ai.ex
@@ -57,13 +57,13 @@ defmodule LangChain.ChatModels.ChatOpenAI do
           {:ok, Message.t() | MessageDelta.t() | [Message.t() | MessageDelta.t()]}
           | {:error, String.t()}

-  @create_fields [:model, :temperature, :frequency_penalty, :n, :stream, :receive_timeout]
+  @create_fields [:endpoint, :model, :temperature, :frequency_penalty, :n, :stream, :receive_timeout]
   @required_fields [:model]

   @spec get_org_id() :: String.t() | nil
@@ -220,7 +220,7 @@ defmodule LangChain.ChatModels.ChatOpenAI do
       Req.new(
         url: openai.endpoint,
         json: for_api(openai, messages, functions),
-        auth: {:bearer, get_api_key()},
+        headers: %{"api-key" => get_api_key()},
         receive_timeout: openai.receive_timeout
       )

```

Hope to make a general base to adapt this change to support more chat models.

cigrainger:
I'd love to support Claude as well. It seems that perhaps `chat_model` could be a behaviour. @brainlid would you be willing to entertain a PR making that change? That would make it easy enough to support Bumblebee a la #26.

brainlid:
@cigrainger I'm just starting to take on Mistral via Bumblebee. Since I haven't done _anything_ with a LLM and Bumblebee, it will help me think through how it might work and what other changes make sense.

brainlid:
I hadn't seen [Claude](https://docs.anthropic.com/claude/docs) before. So many!

chrisbarker:
@brainlid and everyone - I have an implementation for Azure OpenAI that has these and a few other changes to handle streaming responses that contain chunks of json, etc. I thought it might be worth breaking out as it's own chat_model as [python langchain has](https://github.com/langchain-ai/langchain/tree/master/libs/langchain/langchain/chat_models)? 

[...]

PR #60, opened 2023-12-20
feat: add azure openai chat model

chrisbarker:
Based on adapting openai_chat_model to work with azure openai where the content moderation and chat completion api can result in json that is larger than the chunks being returned; requiring a little additional processing before parsing.

The endpoint defines the model being used so no need to specify both model and endpoint. I've kept things simple and put resource endpoint, model and api version all in as the endpoint as together they define the url to use. I have not attempted to break out api versions. note: the current api version I have currently targeted `2023-08-01-preview` is a little behind or is different to the current OpenAI API.

Also it looks like there will likely be a better way to implement this in the codebase in the future, see #28 where @brainlid mentions looking at a behaviour based approach but I'm PR'ing it in case it is useful now for someone and in case there is anything useful here for combining with a behaviour based approach.

### DIFF
--- a/lib/chat_models/chat_azure_openai.ex
+++ b/lib/chat_models/chat_azure_openai.ex
@@ -259,11 +259,6 @@ defmodule LangChain.ChatModels.ChatAzureOpenAI do
   end
 
   def do_api_request(%ChatAzureOpenAI{stream: true} = openai, messages, functions, callback_fn) do
-
-
-    openai |> IO.inspect(label: "openai bahhhh")
-
-
     finch_fun = fn request, finch_request, finch_name, finch_options ->
       resp_fun = fn
         {:status, status}, response ->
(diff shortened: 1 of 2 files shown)


brainlid:
Another contributor created this PR: https://github.com/brainlid/langchain/pull/59

It is around adding Google Gemini support. In the process, he refactored more of the shared code. I'm hoping to take some time over the coming holiday break to dig into it. 

What I would like to improve with this PR is the ability to tell what's actually unique or changed. Extracting out common or reused blocks may help make that clearer. Exactly _how_ to do that is still a question.

chrisbarker:
Just noting most of the conversation around this PR is happening on this issue https://github.com/brainlid/langchain/issues/28

brainlid:
Hi @chrisbarker! I released v0.1.9 which tries to better handle JSON that wraps across delta messages. Could you give this another try and see if it works?

brainlid:
In particular, I'd like to see if the new release supports Azure without needing this additional PR.

brainlid:
I believe we have working Azure support now through #28.

Issue #37, opened 2023-11-13
Update for OpenAI API changes to functions and tools


brainlid:
- [x] The list of functions is deprecated in favor of "tools" - [docs](https://platform.openai.com/docs/api-reference/chat/create#chat-create-functions)
- [x] New `tool_choice` option to add support for. [docs](https://platform.openai.com/docs/api-reference/chat/create#chat-create-tool_choice)
- [x] `function_call` is deprecated in favor of `tool_choice` [docs](https://platform.openai.com/docs/api-reference/chat/create#chat-create-function_call)
- [x] Support receiving multiple function calls from assistant (reported as new ability with gpt-4-preview)

brainlid:
Completed in PR #105

PR #62, opened 2023-12-21
Implement Chat GPT 4 vision (still in beta)

wakeless:
This implements the Chat GPT 4 vision stuff in a very rudimentary way. Adding here mainly as a starting point if others want to use it, or if there's some clear "just do this and we'll merge it" I'm happy to do that as well.

There's a few quirks with the current OpenAI implementation, namely that it can't do the json and functions stuff.


lib/message.ex
```diff
@@ -309,11 +310,23 @@ defmodule LangChain.Message do
...
+
+  def add_image(%Message{role: :user, images: images} = message, image) do
+    %Message{
+      message | images: images ++ [image]
+    }
+  end
+
+  def has_images?(%Message{images: images}) do
+    length(images) > 0
+  end
```
brainlid, line 322: This would be better as two functions and a pattern match.

```elixir
def has_images?(%Message{images: []}), do: false

def has_images?(%Message{images: images}) when is_list(images), do: true
```

This is more explicit and avoids traversing the full list just to know if it's empty or not.

lib/chains/llm_chain.ex
```diff
@@ -317,6 +317,11 @@ defmodule LangChain.Chains.LLMChain do
   """
   @spec add_message(t(), Message.t()) :: t()
   def add_message(%LLMChain{} = chain, %Message{} = new_message) do
+    if Message.has_images?(new_message) and chain.llm.model != "gpt-4-vision-preview" do
```
brainlid, line 320: We don't want model-specific details in LLMChain. We want LLMChain to be model agnostic.

lib/chat_models/chat_open_ai.ex
```diff
@@ -148,11 +152,14 @@ defmodule LangChain.ChatModels.ChatOpenAI do
...
 
-  defp set_response_format(%ChatOpenAI{json_response: true}),
-    do: %{"type" => "json_object"}
+  defp conditionally_add_response_format(output, %ChatOpenAI{model: "gpt-4-vision-preview"}), do: output
+
+  defp conditionally_add_response_format(output, %ChatOpenAI{json_response: true}),
+    do: Map.put(output, :response_format, %{"type" => "json_object"})
+
+  defp conditionally_add_response_format(output, %ChatOpenAI{json_response: false}),
+    do: Map.put(output, :response_format, %{"type" => "text"})
```
brainlid, line 161: Hopefully there's a way to detect images without pinning it to a specific beta model.

lib/message.ex
```diff
@@ -309,11 +310,23 @@ defmodule LangChain.Message do
       do: true
 
   def is_function_call?(%Message{}), do: false
+
+  def add_image(%Message{role: :user, images: images} = message, image) do
```
brainlid, line 314: Adding docs would help. It appears from the tests that this is passing a URL? Not the actual image data, correct?

brainlid (commented):
I've been working with other LLMs that don't support functions and I've been thinking about ways to flag a model as supporting different features. A `supports_images` would be another such flag. That could help remove model-specific awareness from the rest of the library.

lib/chat_models/chat_open_ai.ex
wakeless: Yeah, this is all pretty grose cause it's a beta. TBH, the fact that the beta doesn't support _any_ response_format is :spew:

lib/message.ex
wakeless: Yeah, in this case it is, I'm pretty sure there's an option here to pass base64, so a pattern match on a the image (is_file) or somesuch could handle it easily enough. :+1: on the docs though.

wakeless:
@brainlid Main question I've got is, do you think this is valuable to get in at this point, or better to leave it and wait till it goes to production on openai?

brainlid:
@wakeless First I want to get in the idea of `supports_images` and `supports_functions` in a clean way. I've got some ideas that I want to try. 

With something like that, we could remove the model-specific knowledge leaking out to other parts.

Then I think it would be fine to add.

brainlid:
@wakeless, I just merged in Anthropic support (ie. Claude) for conversations. Claude also supports uploading images as base64 encoded strings directly into a message. 

I'm looking for similarities and differences. Here's what I see so far:

- https://docs.anthropic.com/claude/reference/messages-examples#vision
- https://platform.openai.com/docs/guides/vision

The ChatGPT version supports embedding base64 text (so does Anthropic), but also providing a URL to the image.

They both use a "user" message to provide it, but the content is a list of JSON data, allowing for multiple things to be passed. This allows for text + image.

I'm also thinking about and working on a new Message structure that allows for the ChatGPT "tools" idea. My point is, I think it might be a good time to add support to messages for this type of structure.

brainlid:
Thank you for all of your initial work and interest in this area!

I'm closing this specific PR because it has been implemented on a working branch that is part of issue #37.

To really support multi-modal messages (vision support), I needed to change the Message structure in more significant ways. I also have it working for Anthropic vision. It was very helpful to see 2 different takes on how to do it. Hopefully that future-proofs the implementation for a while longer.

I expect it will be coming soon.

Issue #67, opened 2024-01-06
Request for Community ChatModel: ChatOllama


theaveasso:
I am interested in using the ollama as a self hosted llm model in the Elixir Langchain project. I was wondering if it's possible to extend the support to include ChatModel.ChatOllama as well.

Detail: 
[Python langchain_community.chat_models.ChatOllama](https://api.python.langchain.com/en/stable/chat_models/langchain_community.chat_models.ollama.ChatOllama.html#langchain_community.chat_models.ollama.ChatOllama)

Request:
I kindly request the community's assistance or guidance on how to integrate ChatOllama into the Langchain project. I am new to Elixir, (no idea what I am doing most of the time) but I am eager to learn and contribute to this exciting project.

acalejos:
Correct me if I'm wrong @brainlid , but it seems that you just need to implement Ollama as a struct that implements a `call` function the same as the OpenAI one defined [here](https://github.com/brainlid/langchain/blob/d9fc6a088a53b55d0791252150893a843d9fb66a/lib/chat_models/chat_open_ai.ex#L177) .

There seems to be 2 Ollama Elixir clients [as listed on Hex.pm](https://hex.pm/packages?search=ollama&sort=recent_downloads), but both seem pretty new and in their infancy. You could always roll your own or refer directly to the [Ollama REST API](https://github.com/jmorganca/ollama/blob/main/docs/api.md).  You would use these to hit the API endpoints for chat completion etc. as is done with the OpenAI implementation.

Whether or not the models support `Functions` within the context of Langchain and in the same manner as supported through OpenAIs models are a different problem though, so those parts of the implementation might have to be sorted out differently. 

Once you implement this, then you could simply call the new model as your llm:
```elixir
alias LangChain.Chains.LLMChain
alias LangChain.ChatModels.ChatOpenAI
alias LangChain.Message

{:ok, _updated_chain, response} =
  %{llm: ChatOllama.new!(%{model: "llama2"})}
  |> LLMChain.new!()
  |> LLMChain.add_message(Message.new_user!("Testing, testing!"))
  |> LLMChain.run()

response.content
```

Again, this is just the approach I would take, but having not contributed to this library I could be wrong about the requirements for implementing a new model.

PR #70, opened 2024-01-20
Add support for Ollama open source models

medoror:
Adds support for open source models with [Ollama](https://github.com/jmorganca/ollama) which allows the execution of LLM's locally. Per the current version of the [API](https://github.com/jmorganca/ollama/blob/main/docs/api.md#generate-a-chat-completion) it doesn't seem like function calling is supported yet. See [this issue](https://github.com/jmorganca/ollama/issues/1729) to track on the possible status of function calling

This issue should close https://github.com/brainlid/langchain/issues/67

### DIFF
--- a/lib/chat_models/chat_ollama_ai.ex
+++ b/lib/chat_models/chat_ollama_ai.ex
@@ -212,7 +212,7 @@ defmodule LangChain.ChatModels.ChatOllamaAI do
   Calls the Ollama Chat Completion API struct with configuration, plus
   either a simple message or the list of messages to act as the prompt.
 
-  **NOTE:** This API as of right now does not take any functions. More
+  **NOTE:** This API as of right now does not support functions. More
   information here: https://github.com/jmorganca/ollama/issues/1729
 
   **NOTE:** This function *can* be used directly, but the primary interface
--- a/lib/chat_models/chat_ollama_ai.ex
+++ b/lib/chat_models/chat_ollama_ai.ex
@@ -1,0 +1,370 @@
+defmodule LangChain.ChatModels.ChatOllamaAI do
+  @moduledoc """
+  Represents the [Ollama AI Chat model](https://github.com/jmorganca/ollama/blob/main/docs/api.md#generate-a-chat-completion)
+
+  Parses and validates inputs for making a requests from the Ollama Chat API.
+
+  Converts responses into more specialized `LangChain` data structures.
+
+  The module's functionalities include:
+
+  - Initializing a new `ChatOllamaAI` struct with defaults or specific attributes.
+  - Validating and casting input data to fit the expected schema.
+  - Preparing and sending requests to the Ollama AI service API.
+  - Managing both streaming and non-streaming API responses.
+  - Processing API responses to convert them into suitable message formats.
+
+  The `ChatOllamaAI` struct has fields to configure the AI, including but not limited to:
+
+  - `endpoint`: URL of the Ollama AI service.
+  - `model`: The AI model used, e.g., "llama2:latest".
+  - `receive_timeout`: Max wait time for AI service responses.
+  - `temperature`: Influences the AI's response creativity.
+
+  For detailed info on on all other parameters see documentation here:
+  https://github.com/jmorganca/ollama/blob/main/docs/modelfile.md#valid-parameters-and-values
+
+  This module is for use within LangChain and follows the `ChatModel` behavior,
+  outlining callbacks AI chat models must implement.
+
+  Usage examples and more details are in the LangChain documentation or the
+  module's function docs.
+  """
+  use Ecto.Schema
+  require Logger
+  import Ecto.Changeset
+  alias __MODULE__
+  alias LangChain.ChatModels.ChatModel
+  alias LangChain.Message
+  alias LangChain.MessageDelta
+  alias LangChain.LangChainError
+  alias LangChain.ForOpenAIApi
+  alias LangChain.Utils
+
+  @behaviour ChatModel
+
+  @type t :: %ChatOllamaAI{}
+
+  @create_fields [
+    :endpoint,
+    :mirostat,
+    :mirostat_eta,
+    :mirostat_tau,
+    :model,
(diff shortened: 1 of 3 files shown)


brainlid (changes requested):
Thanks for doing all the work implement this! There are a few changes needed before merging it.

- Please refer to [this PR](https://github.com/brainlid/langchain/pull/68) where retries for failed Mint connections was added. Req is trying to handle these, but it's currently not enough. The PR does library level retries when we detect that the Mint connection pulled from the pool is already closed and triggering a retry.
- Having module level docs and docs on important public functions is really helpful for other developers. For the module doc, assume the reader doesn't know what ChatOllamaAI is at all. So a brief overview and links to other external relevant docs is helpful.

lib/chat_models/ollama_ai_fields.ex
```diff
@@ -0,0 +1,78 @@
+defmodule LangChain.ChatModels.OllamaAIFields do
+  defmacro fields do
```
brainlid, line 2: Why do this in a macro and a separate file?

test/chat_models/chat_ollama_ai_test.exs
```diff
@@ -0,0 +1,224 @@
...
+    end
+
+    test "generates a map for an API call with user and system messages", %{ollama_ai: ollama_ai} do
+      user_message = "What color is the sky?"
+      system_message = "You are a weather man"
+
+      data =
+        ChatOllamaAI.for_api(
+          ollama_ai,
+          [Message.new_user!(user_message), Message.new_system!(system_message)],
```
brainlid, line 104: Is this intentionally in reverse order? Typically the system message is expected to be first and user messages follow.

Instead of using procedural `get_in` calls with Access.at/1 calls, pattern matches are more clear.  For example:

```elixir
assert [system_msg, user_msg] = data.messages
assert system_msg.role == :system
```
This fetches the messages and asserts a list of 2 elements all at once.

Or even like this: 
```elixir
assert [%Message{role: :system} = system_msg, %Message{role: :user} = user_msg] = data.messages
assert system_msg.content == "You are a weather man"
assert user_msg.content == "What color is the sky?"
```

Where we can assert 2 messages received in a particular order and the roles are asserted at that time. Then the longer content can be asserted separately.

lib/chat_models/ollama_ai_fields.ex
```diff
@@ -0,0 +1,78 @@
...
+    quote do
+      @receive_timeout 60_000 * 5
+
+      field :endpoint, :string, default: "http://localhost:11434/api/chat"
```
brainlid, line 6: Is there any support for overriding this endpoint? This default value should be explained in the ChatOllamaAI's module doc along with information or an example of how to override.

lib/chat_models/ollama_ai_fields.ex
```diff
@@ -0,0 +1,78 @@
+defmodule LangChain.ChatModels.OllamaAIFields do
+  defmacro fields do
+    quote do
+      @receive_timeout 60_000 * 5
```
brainlid, line 4: Is this really a 5 minute timeout? Why so long? Is it really slow to run these models locally? We might want to have the ability to override the timeout in that case.

lib/chat_models/ollama_ai_fields.ex
```diff
@@ -0,0 +1,78 @@
...
+
+      field :endpoint, :string, default: "http://localhost:11434/api/chat"
+
+      # Enable Mirostat sampling for controlling perplexity.
+      # (default: 0, 0 = disabled, 1 = Mirostat, 2 = Mirostat 2.0)
+      field :mirostat, :integer, default: 0
```
brainlid, line 10: For all these fields with special docs, it would be helpful to include links to the docs where these come from.

lib/chat_models/ollama_ai_fields.ex
medoror: With so many fields, the chat_ollama_ai.ex was getting hard to read. I can inline the fields back in though if that is your preference. If you have any alternatives for DRY'ing those fields up let me know

test/chat_models/chat_ollama_ai_test.exs
medoror: This makes sense. I am still pretty new to elixir so this is helpful!

lib/chat_models/ollama_ai_fields.ex
medoror: Will make a note in the module docs on this.  The app within the [image exposes port 11434](https://github.com/jmorganca/ollama/blob/main/Dockerfile#L119) but their is a world where a client may want to deploy the image with a different port mapping thus requiring an override

lib/chat_models/ollama_ai_fields.ex
medoror: No clue why. The only information I found about time outs was here: https://github.com/jmorganca/ollama/pull/1257

Client should be able to override because it is a valid field in the struct

medoror:
@brainlid Addressed those comments. Let me know if I'm missing anything else

lib/chat_models/ollama_ai_fields.ex
brainlid: Thank, I see you already did. :+1:

lib/chat_models/chat_ollama_ai.ex
```diff
@@ -0,0 +1,370 @@
...
+      top_k: model.top_k,
+      top_p: model.top_p
+    }
+  end
+
+  @doc """
+  Calls the Ollama Chat Completion API struct with configuration, plus
+  either a simple message or the list of messages to act as the prompt.
+
+  **NOTE:** This API as of right now does not take any functions. More
```
brainlid, line 215:
```suggestion
  **NOTE:** This API as of right now does not support functions. More
```

brainlid (approved):
Minor comment tweak suggested. Looks good!

brainlid:
Thanks for the work and your patience!

❤️💛💙💜

Issue #71, opened 2024-02-05
`LLMChain.run()` for ollama doesn't work


petrkozorezov:
```elixir
alias LangChain.{Chains.LLMChain, Message, LangChain.ChatModels.ChatOllamaAI}
LLMChain.new!(%{
  llm: LangChain.ChatModels.ChatOllamaAI.new!(%{
    model: "mixtral:8x7b-instruct-v0.1-q4_K_M"
  }),
  verbose: true,
})
|> LLMChain.add_message(Message.new_user!("hello world!"))
|> LLMChain.run()
```

```
** (CaseClauseError) no case clause matching: {:ok, %LangChain.Message{content: " Hello! It's nice to see you. Is there something specific you would like to talk about or ask me a question? I'm here to help with any programming-related questions you have.\n\nIf you don't have anything specific in mind, that's okay too! We can just chat about whatever interests you. Do you enjoy coding? What are some of your favorite programming languages and why do you like them?\n\nI'm particularly fond of Python, because it's a great language for beginners to learn, with a clean and easy-to-understand syntax. It's also very versatile and can be used for a wide variety of applications, from web development to data analysis to machine learning.\n\nBut enough about me! I want to hear about you. What brings you to coding? Is there something specific you're hoping to learn or accomplish with your programming skills? Let me know and I'll do my best to help you out.", index: nil, status: :complete, role: :assistant, function_name: nil, arguments: nil}}
    (langchain 0.1.7) lib/chains/llm_chain.ex:209: LangChain.Chains.LLMChain.do_run/1
    (langchain 0.1.7) lib/chains/llm_chain.ex:170: LangChain.Chains.LLMChain.run/2
    iex:4: (file)
```

brainlid:
Ah, that must be from the main branch.

@medoror, can you please help diagnose and verify this?

medoror:
yea I think I noticed this. Will report back later today on my findings

petrkozorezov:
I think the problem is [here](https://github.com/brainlid/langchain/blob/ba2b8d3db7e46b805279c158b200759f0bda7401/lib/chains/llm_chain.ex#L203). This case doesn't handle case `{:ok, message}`, but according to [spec](https://github.com/brainlid/langchain/blob/ba2b8d3db7e46b805279c158b200759f0bda7401/lib/chat_models/chat_model.ex#L8) it's possible.

medoror:
[...]

PR #72, opened 2024-02-06
Add clause to match call_response spec

medoror:
Adds a clause to match an `{:ok, Message.t()}` payload according to the [spec for the call function](https://github.com/brainlid/langchain/blob/main/lib/chat_models/chat_model.ex#L2)

Addresses #71


brainlid:
Thanks @medoror! Sorry for the delay, been busy in other areas of life. I considered going a couple different directions with this, but ultimately opted for your direct solution. It's clean too.

I was wondering if the ChatModel should always return a message in a list, just to simplify and consolidate the return types. Then I wondered if the OpenAI implementation should actually NOT return a list. The only reason it does is because a caller can request "n" variations which come back in an array. I see no practical use for that though.

In the end, both of the other changes would be more invasive and dramatic and I'm not entirely sure yet which is the better way to go. So this solution works great. Thanks!

❤️💛💙💜

PR #73, opened 2024-02-09
Add max_tokens option for OpenAI calls

matthusby:
One of the optional fields that the OpenAI Api allows for in chat completions is `max_tokens`.

https://platform.openai.com/docs/api-reference/chat/create#chat-create-max_tokens


brainlid:
Nice @matthusby! Thanks!

❤️💛💙💜

Issue #83, opened 2024-03-23
Explore/add Anthropic API support (Cluade AI)


brainlid:

brainlid:
Basic Anthropic support added. They don't have an official API for tools, so functions are not implemented at this point.

youfun:
anthropic releases   the tools ，is that what you say “an official API for tools”：

see the doc：
https://docs.anthropic.com/claude/docs/tool-use

Tool use (function calling)
Claude is capable of interacting with external client-side tools and functions, allowing you to equip Claude with your own custom tools to perform a wider variety of tasks.

🎉
Tool use public beta

We're excited to announce that tool use is now in public beta! To access this feature, you'll need to include the anthropic-beta: tools-2024-04-04 header in your API requests.

We'll be iterating on this open beta over the coming weeks, so we appreciate all your feedback. Please share your ideas and suggestions using this [form](https://forms.gle/BFnYc6iCkWoRzFgk7).

Here's an example of how to provide tools to Claude using the Messages API:

Shell
Python

curl https://api.anthropic.com/v1/messages \
  -H "content-type: application/json" \
  -H "x-api-key: $ANTHROPIC_API_KEY" \
  -H "anthropic-version: 2023-06-01" \
  -H "anthropic-beta: tools-2024-04-04" \
  -d '{
    "model": "claude-3-opus-20240229",
    "max_tokens": 1024,
    "tools": [
      {
        "name": "get_weather",
        "description": "Get the current weather in a given location",
        "input_schema": {
          "type": "object",
          "properties": {
            "location": {
              "type": "string",
              "description": "The city and state, e.g. San Francisco, CA"
            }
          },
          "required": ["location"]
        }
      }
    ],
    "messages": [
      {
        "role": "user",
        "content": "What is the weather like in San Francisco?"
      }
    ]
  }'
Please note that during the beta period:

Streaming (stream=true) is not yet supported. We plan to add streaming support in a future beta version.
While the feature is production-ready, we may introduce multiple beta versions before the final release.
Tool use is not yet available on third-party platforms like Vertex AI or AWS Bedrock, but is coming soon.

brainlid:
Correct. See issue #94

PR #86, opened 2024-03-24
Add support for Anthropic

benjreinhart:
This adds support for [Anthropic's "create message" endpoint](https://docs.anthropic.com/claude/reference/messages_post) which supports the new [Claude 3 models](https://docs.anthropic.com/claude/docs/models-overview#model-recommendations).

Addresses #83 

### Notes

* Anthropic supports a system message, but it is not included in the list of messages. It is its own top-level `system` attribute.
* The SSE streaming includes a bit more / different data than the other providers. This includes another line containing: `event: <event_name>`. Additionally, there are some events that contain no useful information, so they must be ignored. See [streaming](https://docs.anthropic.com/claude/reference/messages-streaming#raw-http-stream-response).

### Example

(pulled from the livebook)

```elixir
alias LangChain.Message
alias LangChain.MessageDelta
alias LangChain.ChatModels.ChatAnthropic

key = "<ANTHROPIC_API_KEY>"

callback = fn
  %MessageDelta{} = data ->
    # we received a piece of data
    IO.write(data.content)

  %Message{} = data ->
    # we received the finshed message once fully complete
    IO.puts("")
    IO.puts("")
    IO.inspect(data.content, label: "COMPLETED MESSAGE")
end

{:ok, chain, response} = (%{llm: ChatAnthropic.new!(%{api_key: key, stream: true})}
  |> LLMChain.new!()
  |> LLMChain.add_messages([
    Message.new_system!("You are a helpful assistant."),
    Message.new_user!("Write a haiku about the capital of the United States")
  ])
  |> LLMChain.run(callback_fn: callback))
```


brainlid:
Awesome! Thanks for doing all the initial groundwork! I setup an account recently, so now it's just testing and reviewing! :boom:

lib/chat_models/chat_anthropic.ex
```diff
@@ -0,0 +1,560 @@
...
+      status: stop_reason_to_status(stop_reason)
+    }
+    |> MessageDelta.new()
+    |> to_response()
+  end
+
+  # These streaming events don't contain any useful data to return to callers. Ignore.
+  def do_process_response(%{"type" => type})
+      when type in ["message_start", "ping", "content_block_stop", "message_stop"] do
+    :ignore
```
brainlid, line 438: Is this being used? It's not part of the spec and doesn't really fit in with the other return types. Wondering if it's actively being used here or if there's another way to convey this information.

lib/chat_models/chat_anthropic.ex
```diff
@@ -0,0 +1,560 @@
...
+    #     data: {"type": "content_block_stop", "index": 0}
+    #
+    #     event: message_delta
+    #     data: {"type": "message_delta", "delta": {"stop_reason": "end_turn", "stop_sequence":null, "usage":{"output_tokens": 15}}}
+    #
+    #     event: message_stop
+    #     data: {"type": "message_stop"}
+    #
+    #
+    # See Anthropic documentation: https://docs.anthropic.com/claude/reference/messages-streaming
```
brainlid, line 499: Thanks for documenting this with references as well!

lib/chat_models/chat_anthropic.ex
```diff
@@ -0,0 +1,560 @@
...
+
+    # This captures `event` and `data`
+    @event_data_re ~r/^\s*event:\s*(?<event>\S*)\s*data:\s*(?<data>[^\n]+)\s*$/
```
brainlid, line 502: I'm wondering if we could do binary pattern matches on the event type and data instead of RegEx?

We'd need a catch-all to future proof from news events being added in the future.

As an example, it might look like this:

```elixir
def decode("event: message_start\ndata: " <> data) do
  # process data
end

def decode("event: ping\n" <> _rest), do: :ignore

# ... more events we handle or ignore ...

def decode("event: " <> rest) do
  Logger.error("Unsupported event received")
  :ignore
end
```

Something like this is very fast to execute but, more importantly, more clear and explicit.

You know more about this protocol. Is this a possible solution?

brainlid (commented):
Impressive work! Have some questions.

lib/chat_models/chat_anthropic.ex
benjreinhart: With the way the code is currently organized, yes this clause would be needed when streaming responses.

Here's a curl request with its response you can use to see these values:

```shell
curl https://api.anthropic.com/v1/messages \
--header "x-api-key:  $ANTHROPIC_API_KEY" \
--header "content-type: application/json" \
--header "anthropic-version: 2023-06-01" \
--data \
'{
    "model": "claude-3-opus-20240229",
    "max_tokens": 1024,
    "messages": [
        {"role": "user", "content": "Hello, world"}
    ],
    "stream": true
}'
```

```
event: message_start
data: {"type":"message_start","message":{"id":"msg_01F8kFTQzuVBK3V53N36vZGN","type":"message","role":"assistant","content":[],"model":"claude-3-opus-20240229","stop_reason":null,"stop_sequence":null,"usage":{"input_tokens":10,"output_tokens":1}}}

event: content_block_start
data: {"type":"content_block_start","index":0,"content_block":{"type":"text","text":""}               }

event: ping
data: {"type": "ping"}

event: content_block_delta
data: {"type":"content_block_delta","index":0,"delta":{"type":"text_delta","text":"Hello"} }

event: content_block_delta
data: {"type":"content_block_delta","index":0,"delta":{"type":"text_delta","text":"!"} }

event: content_block_delta
data: {"type":"content_block_delta","index":0,"delta":{"type":"text_delta","text":" How"}              }

event: content_block_delta
data: {"type":"content_block_delta","index":0,"delta":{"type":"text_delta","text":" can"}      }

event: content_block_delta
data: {"type":"content_block_delta","index":0,"delta":{"type":"text_delta","text":" I"}}

event: content_block_delta
data: {"type":"content_block_delta","index":0,"delta":{"type":"text_delta","text":" assist"}         }

event: content_block_delta
data: {"type":"content_block_delta","index":0,"delta":{"type":"text_delta","text":" you"}           }

event: content_block_delta
data: {"type":"content_block_delta","index":0,"delta":{"type":"text_delta","text":" today"}  }

event: content_block_delta
data: {"type":"content_block_delta","index":0,"delta":{"type":"text_delta","text":"?"}    }

event: content_block_stop
data: {"type":"content_block_stop","index":0         }

event: message_delta
data: {"type":"message_delta","delta":{"stop_reason":"end_turn","stop_sequence":null},"usage":{"output_tokens":12}     }

event: message_stop
data: {"type":"message_stop"    }

```

Now, we could change the parser logic below to try and ignore these events rather than allowing them to bubble up to this function, but to me that felt like an awkward abstraction because it would mean that the code responsible for string parsing is also aware of what events have significance to the caller. The original intent was to have the string parsing be its own concern and leave any high-level filtering logic up to callers.

That said, this `:ignore` pattern here is less than ideal so I don't feel strongly. Happy to rework a bit

lib/chat_models/chat_anthropic.ex
benjreinhart: So I had done something like this in my own code base with OpenAI originally and I will say that there have been a few times where we noticed an extra or a missing whitespace/newline which broke our parser. That seems to break their contract, so to be fair I assume that is a bug on their end. While that was openai, I wouldn't be surprised to see the same thing happen with other APIs. The goal of the regex was to be a bit less sensitive while parsing.

I agree the pattern matching solution reads much nicer though so I can switch to that

lib/chat_models/chat_anthropic.ex
brainlid: Thanks for putting that all into context here. With the event information always included in the data as "type", is there any reason to pay attention to the event on the outside? 

Wondering if we always grab the data and parse it to JSON, and ingore the "event" part of the message. Wouldn't that simplify the parsing? Is there another reason to parse the the outer event?

benjreinhart: At the moment, no reason to keep the event that I can think of. My line of thinking was to keep the parser pure in the sense that it just returns objects representing the raw stream, but that's just a preference. Since there's no real need, I removed those.

benjreinhart:
@brainlid I think I addressed feedback but let me know if you'd prefer any other changes, happy to keep whatever style you prefer.

One other thing of note, I am not filtering out the `content_block_start` since it does contain a field indicating there might be text in there. However, in all my testing, it's always empty. This may be something we want to filter out but it's unclear to me whether it's guaranteed to always be empty or if may occasionally have a legitimate value in it. Let me know what you think.

brainlid:
Thanks @benjreinhart! Just letting you know what I'm doing now. I haven't forgotten this. :slightly_smiling_face: 

I agree with you on keeping the `content_block_start` event.

Anthropic is just the 3rd example of a different way to stream AI responses. Most of them matched up with OpenAI's structure. Rather than saying, "Well, what if you try it this way..." over and over, I decided to play with it myself.

I'm working off of your fork. I hope to have something ready soon-ish that goes a little further with some internal refactoring.

brainlid:
@benjreinhart, you had a test case for handling when a streamed response was cut-off. Specifically, when the event name is split across lines. My refactor breaks this and I'm wondering if that's something you've actually observed and is expected to happen, or if you were just checking that it could be handled?

```
"event: content_blo"
...
"ck_delta\ndata: {\"type\":\"content_block_delta\",\"index\":0,\"de"
```

benjreinhart:
@brainlid I am not sure tbh. I don't know enough about networks and what guarantees can be made with the arrival of "complete" events from SSEs, so I'd probably err on the side of caution. I _think_ it's likely that incomplete events could arrive even if the server originally sent a whole event (eg, proxies buffering, clients buffering, or just network instability).

It looks like you found an example in the wild, but even if not, I'd personally try to be robust to this type of thing unless I can get guarantees that it won't happen.

brainlid:
@benjreinhart I've got it working well and ready for testing. I'll push the branch to the main repo as I'm having a hard time pushing it as a PR to your branch. :man_shrugging: 

It's fast and works well. I'd love to hear your experience with it. I'm bummed that Anthropic doesn't support tools/functions. How are you integrating with it? Just curious.

brainlid:
@benjreinhart can you enable me to push changes to your branch? That seems to be the cleanest way and keeps you having credit for your work. 

https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork

benjreinhart:
It looks like it is enabled 

Screenshot 2024-04-03 at 6 12 20 AM

brainlid:
Ok! I just had to figure out how to do it. I pushed my refactor to your fork/anthropic branch. Try it out and let me know if you're good to go forward with it.

brainlid:
@benjreinhart Here's a breakdown of the main changes:

- separate decode step for gathering a complete set of data to process. The buffering is handled there.
- the decode step converts to JSON
- separate step for processing the JSON into structs
- all models updated to use multi-step process
- refactored the OpenAI specific stream format to be in the model's module
- other models reference the OpenAI version explicitly if they share/reuse the same format
- fixed some type spec inconsistencies
- LLMChain can handle receiving a list of MessageDeltas that aren't nested inside "N" choice sub-lists

What's left:
- I want to do a refactor/rewrite of the OpenAI decode step to use streams as well. Will be a separate change.

lib/chat_models/chat_anthropic.ex
```diff
@@ -455,107 +437,72 @@ defmodule LangChain.ChatModels.ChatAnthropic do
...
+        {Enum.slice(entries, 0..-2//1), List.last(entries)}
+      end
 
-    defp decode_event("event: content_block_stop\n" <> _rest) do
-      :ignore
-    end
+    processed =
+      to_process
+      # Trim whitespace from each line
+      |> Stream.map(&String.trim/1)
```
benjreinhart, line 459: Was the thinking behind this implementation over the previous one using pattern matching to be more robust to whitespace?

benjreinhart (commented):
LGTM! I'm not opinionated either way, but wondering the benefits of the stream implementation here? I think the lists the streams are iterating over are always small (1-5 element lists per chunk) and we will always iterate over every element, I think?

benjreinhart:
> I'm bummed that Anthropic doesn't support tools/functions. How are you integrating with it?

I haven't used functions with Anthropic, and my understanding was that they did not support any first-class API for that. However, I recently stumbled upon something in the docs that seems to indicate they have some support for it: https://docs.anthropic.com/claude/docs/functions-external-tools

lib/chat_models/chat_anthropic.ex
brainlid: I wasn't sure how large the chunks might be, but it felt like Streams were a good fit. I've profiled Streams vs Enum approaches before and find them very comparable in performance on the small end. Streams have the greatest benefit when the data is very large.

Really, my goal was to may it easier to read and reason about and to make pattern matching more heavily featured.

brainlid:
@benjreinhart 

> However, I recently stumbled upon something in the docs that seems to indicate they have some support for it: https://docs.anthropic.com/claude/docs/functions-external-tools

Excellent! Their doc says official support is "coming soon". I was looking at the JS LangChain approach to function support with Claude, as I think it will work well for Bumblebee as well... which I want to do anyway.

I didn't know about that resource, so thanks for highlighting it!

brainlid:
Thanks for your awesome help on this!
❤️💛💙💜

PR #93, opened 2024-04-03
Azure OpenAI support: send the auth api key differently

brainlid:
- sends it both ways
- Azure OpenAI requires an "api-key" header
- Standard OpenAI requires auth bearer header
- formatting


brainlid:
Special thanks to @raulchedrese for identifying the minimum required change for adding support.
❤️💛💙💜

brainlid:
Closes #28 if everything checks out.

PR #98, opened 2024-04-06
Wrap Anthropic message when processing response

yujonglee:
This line cause problem for current Anthropic implementation:

https://github.com/brainlid/langchain/blob/1dab891c48d26f0127991c784e6a2bb7f1050188/lib/utils.ex#L109

This PR fix it by wrapping message.

### DIFF
--- a/lib/chains/llm_chain.ex
+++ b/lib/chains/llm_chain.ex
@@ -225,7 +225,6 @@ defmodule LangChain.Chains.LLMChain do
           do: IO.inspect(updated_chain.last_message, label: "COMBINED DELTA MESSAGE RESPONSE")
 
         {:ok, updated_chain}
-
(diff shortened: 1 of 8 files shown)


brainlid:
Thanks! I like the fix. Can you update the PR to not include the formatting changes? Yes, they are valid formatting changes, however, they are not related to the 1 line fix.

yujonglee:
@brainlid makes sense. just did that!

brainlid:
Thanks! 
❤️💛💙💜

I ended up solving it a different way by updating `Utils.fire_callback`. The spec for `fire_callback` should allow for a non-list, and I think that's right. I didn't want converting a single message to return a list of 1 messages either.

But this pointed out that I was missing some test coverage, which I've also addressed.

PR #123, opened 2024-05-24
Fix elixir 1.17 warnings

MrYawe:
Fixes
```
warning: 0..-2 has a default step of -1, please write 0..-2//-1 instead
  test/chat_models/chat_ollama_ai_test.exs:160: ChatModels.ChatOllamaAITest."test call/2 basic content example with streaming"/1
```


brainlid (approved):
Thanks for fixing this!

PR #124, opened 2024-05-29
Add ChatVertexAI

raulchedrese:
Here is my first crack at a a ChatVertexAI module. I'm currently using this module in production with good results.

I adapted the tests from the existing ChatGoogleAI module.

I also have a fix in here for an issue I ran into when using OpenAI through Azure. I don't know that this is an ideal way to deal with the issue though. Let me know if you want me to remove that fix from this PR.


lib/chains/llm_chain.ex
```diff
@@ -226,6 +226,18 @@ defmodule LangChain.Chains.LLMChain do
 
         {:ok, updated_chain}
 
+      # When calling the latest OpenAI models through Azure it sometimes returns an empty list
+      # as the first element of a list of MessageDeltas. This is a workaround to handle that.
+      # It would likely be better to handle this in ChatOpenAI instead.
+      {:ok, [[] | _] = deltas} ->
```
brainlid, line 232: Yes, this peculiarity should be handled in the LLM module instead. It would also need a test case. 

Perhaps we can pull this out into a separate PR?

brainlid (changes requested):
Just looking over the ChatVertextAI module looks good. Thanks!

brainlid (approved):
Thanks for helping to finalize the changes! Let's try this out let others bang on it too.

Thanks!
❤️💛💙💜

PR #127, opened 2024-05-31
Handle data extraction where tool_calls response is a list.

cpursley:
A high percentage of the responses for the data extraction chain have been returning a list of tool_calls instead of a single result in the "info". This PR maps "info" when there are multiple tool calls. Not sure how to set up a test case for that but open to ideas.

### DIFF
--- a/lib/chains/data_extraction_chain.ex
+++ b/lib/chains/data_extraction_chain.ex
@@ -105,23 +105,7 @@ Passage:
       |> LLMChain.add_tools(build_extract_function(json_schema))
       |> LLMChain.add_messages(messages)
       |> LLMChain.run()
-      |> case do
-        {:ok, _updated_chain,
-         %Message{
-           role: :assistant,
-           tool_calls: [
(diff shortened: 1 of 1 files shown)


brainlid:
@cpursley yes, the data extraction chain needs to be fixed up. I'm working through a big change right now that will make the data extraction process easier particularly for models that don't support tool calls (like Bumblebee self-hosted models).

Because every model is so unique in terms of what it can do and the prompts needed for a successful extraction, I'm less and less convinced that a standard one-size-fits-all chain can work well. It might be more of, "Here's a guide and a template for building your own data extraction process".

I'm hopefully finishing up the code changes now and will write more about this.

cpursley:
This also might be that I'm using Llama but via an OpenAI compatible API fwiw.

Sounds good, let me know how I can help.

brainlid:
This may still be an issue with the new version. I'm not sure. You'll have to give it a try and let me know what's happening.

PR #135, opened 2024-06-11
Added a failing test for tool usage with Gemini

nileshtrivedi:
I have simply copied the OpenAI calculator test for Gemini and I'm able to run it with `mix test test/tools/calculator_gemini_test.exs --include live_call --include live_gemini`, having first defined the Gemini API Key in a shell variable with: `export GEMINI_API_KEY=your_gemini_api_key`.

This test now fails:

```
  1) test live test performs repeated calls until complete with a live LLM (LangChain.Tools.CalculatorTest)
     test/tools/calculator_gemini_test.exs:68
     ** (MatchError) no match of right hand side value: {:error, %LangChain.Chains.LLMChain{llm: %LangChain.ChatModels.ChatGoogleAI{endpoint: "https://generativelanguage.googleapis.com/v1beta", version: "v1beta", model: "gemini-pro", api_key: nil, temperature: 0.0, top_p: 1.0, top_k: 1.0, receive_timeout: 60000, stream: false}, verbose: false, verbose_deltas: false, tools: [%LangChain.Function{name: "calculator", description: "Perform basic math calculations or expressions", display_text: nil, function: #Function<0.76765322/2 in LangChain.Tools.Calculator.execute>, async: true, parameters_schema: %{type: "object", required: ["expression"], properties: %{expression: %{type: "string", description: "A simple mathematical expression"}}}, parameters: []}], _tool_map: %{"calculator" => %LangChain.Function{name: "calculator", description: "Perform basic math calculations or expressions", display_text: nil, function: #Function<0.76765322/2 in LangChain.Tools.Calculator.execute>, async: true, parameters_schema: %{type: "object", required: ["expression"], properties: %{expression: %{type: "string", description: "A simple mathematical expression"}}}, parameters: []}}, messages: [%LangChain.Message{content: "Answer the following math question: What is 100 + 300 - 200?", processed_content: nil, index: nil, status: :complete, role: :user, name: nil, tool_calls: [], tool_results: nil}], custom_context: nil, message_processors: [], max_retry_count: 3, current_failure_count: 0, delta: nil, last_message: %LangChain.Message{content: "Answer the following math question: What is 100 + 300 - 200?", processed_content: nil, index: nil, status: :complete, role: :user, name: nil, tool_calls: [], tool_results: nil}, needs_response: true, callbacks: [%{on_tool_response_created: #Function<2.77608786/2 in LangChain.Tools.CalculatorTest."test live test performs repeated calls until complete with a live LLM"/1>}]}, "Unexpected response"}
     code: {:ok, updated_chain, %Message{} = message} =
     stacktrace:
       test/tools/calculator_gemini_test.exs:85: (test)

     The following output was logged:
     
     18:03:13.653 [error] Trying to process an unexpected response. ""
     
     18:03:13.653 [error] Error during chat call. Reason: "Unexpected response"
```


brainlid:
Thanks @nileshtrivedi for testing it out with GoogleGemini! Looks like I'll have to get an account. I can't run this currently. 

BTW, a repeatable way to work with a service like this is to IO.inspect the received data before it's processed (and failing). Then setup a fake API response that returns that data. Then it's easy to run and re-run the development without hitting a live service. This is especially helpful when it works one time but not the other because of how the LLM varies a response.

Just letting you know, I won't merge in a new test file for Gemini and would rather create a Gemini specific test in the existing test file.

PR #141, opened 2024-06-21
WIP: Add LLMChain.set_tool_choice

stevehodgkiss:
This PR adds an option to specify `tool_choice`. I found that a lot of structured data extraction responses contained `multi_tool_use.parallel`. Specifying a tool via `tool_choice` avoids having to deal with this, and forces the use of the given tool.

I've only added this to the OpenAI adapter so far. Thoughts?

### DIFF
--- a/lib/chains/llm_chain.ex
+++ b/lib/chains/llm_chain.ex
@@ -194,6 +194,10 @@ defmodule LangChain.Chains.LLMChain do
     |> apply_action!(:update)
   end
 
+  @doc """
+  Set the tool_choice for the LLMChain.
+  """
+  @spec set_tool_choice(t(), ChatModel.tool_choice()) :: t() | no_return()
   def set_tool_choice(%LLMChain{} = chain, tool_choice) do
     chain
     |> change()
--- a/lib/chains/llm_chain.ex
+++ b/lib/chains/llm_chain.ex
@@ -65,6 +65,7 @@ defmodule LangChain.Chains.LLMChain do
     # verbosely log each delta message.
     field :verbose_deltas, :boolean, default: false
     field :tools, {:array, :any}, default: [], virtual: true
+    field :tool_choice, :string, default: nil, virtual: true
     # set and managed privately through tools
     field :_tool_map, :map, default: %{}, virtual: true
 
@@ -190,6 +191,13 @@ defmodule LangChain.Chains.LLMChain do
     |> change()
     |> cast(%{tools: updated}, [:tools])
     |> build_tools_map_from_tools()
+    |> apply_action!(:update)
+  end
+
+  def set_tool_choice(%LLMChain{} = chain, tool_choice) do
+    chain
+    |> change()
+    |> cast(%{tool_choice: tool_choice}, [:tool_choice])
     |> apply_action!(:update)
   end
 
@@ -343,7 +351,7 @@ defmodule LangChain.Chains.LLMChain do
     %module{} = chain.llm
 
     # handle and output response
-    case module.call(chain.llm, chain.messages, chain.tools) do
+    case module.call(chain.llm, chain.messages, chain.tools, chain.tool_choice) do
       {:ok, [%Message{} = message]} ->
         if chain.verbose, do: IO.inspect(message, label: "SINGLE MESSAGE RESPONSE")
         {:ok, process_message(chain, message)}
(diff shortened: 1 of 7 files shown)


avergin:
Just opened a [PR](https://github.com/brainlid/langchain/pull/142) trying to resolve the same issue.

I've added the `tool_choice` attributes to each model struct (ChatOpenAI and ChatAnthropic) to map them specifically to each relevant API spec, rather than a global attribute inside LLMChain.

My thinking for putting `tool_choice` inside each related model struct was to be able to map internal functionality of the relevant models as much as possible to make it more future proof. But I don't know if this is the right way to tackle the problem. Any thoughts on this?

stevehodgkiss:
Hi @avergin 👋 

The modelling of the ToolChoice in your PR is definitely more fully featured than this one currently (i.e. the ability to set `none`/`auto` in addition to the specific tool). This change currently only lets you define the tool to use, although it could be improved to allow defining `none`/`auto` etc.

> My thinking for putting tool_choice inside each related model struct was to be able to map internal functionality of the relevant models as much as possible to make it more future proof. But I don't know if this is the right way to tackle the problem. Any thoughts on this?

I think the `tool_choice` should be able to be specified/changed at the same points that the `tools` can be changed. I'm using the changes in this PR to extract structured data and under certain conditions extend the initial message thread (question/tool result messages) with another tool/extraction (so the LLM has the context of the first answer), and I need to be able to change the tool choice as well as set the new tool at that point.

The change in this PR requires a change to the adapter interface (an extra arg), which I why I wanted to get some feedback on the approach from @brainlid before going through the other adapters to update them / add tool choice to the ChatAnthropic adapter.

brainlid:
Thank you @stevehodgkiss for the contribution! I agree with @avergin that this particular setting belongs on the model, especially since each model may implement it differently or not at all.

stevehodgkiss:
Hi @brainlid 👋 

Each model may implement tools differently or not at all, but they're not defined on the adapter/chat model 😄 

Having it on the chat model would mean the adapter would need to be changed out mid thread if you want to change the tools/tool choice exposed in the next message (`chain = %LLMChain{chain | llm: llm}`), which doesn't seem ideal. I see the adapter/chat model as something to be configured once per chain, not changing as tool choice needs change.

One idea to get around the fact that these are all adapter/chat-model specific could be to have a pass-through map, where the user defines the tool_choice in the format the model in use expects - `LLMChain.set_tool_choice(%{"type" => "tool", "name" => "greet"}}`. There might be better ways of modelling this though 🤔

brainlid:
Thanks @stevehodgkiss for sharing more. I'd like to understand a bit better. I'm trying to imagine a scenario where I'd want to force a tool choice later in the process. Can you walk me through this? What model are you using too? 

Here's what I'm wondering. I totally get the idea of forcing a tool call with the first use, particularly when doing data extraction or route selection. But how would I know part way through a chat sequence that NOW I should force a specific tool call? If that's the case, my user message could say, "Use tool X to process blah". I suppose that depends on the model being used if it can handle it or not. 

My other thought comes from other LangChain uses I've seen. Like collapsing the past messages into a single prompt of chat history and issuing a new chain/chat/command to perform the tool call based on the previous exchange.

Also, the messages in chain are portable and could be moved to a new chain/model to run in unique circumstances.

I'm just not seeing when or why I'd change it part way through. Can you help me see what you're seeing?

stevehodgkiss:
The use case for this is extracting structured data (gpt4o). An initial extraction call can return answers/citations that are ambiguous (multiple of the same answer in a given section/chunk), and a subsequent tool call is used to disambiguate them. Each word can be mapped to a bounding box on a page, so it's important to get the correct occurrence of a given answer. There might be better ways of doing this, such as asking for a larger substring quote in addition to the answer to reduce the chances of ambiguity, but this works really well as a fallback.

Having `tool_choice` on the chat models and `tools` on LLMChain feels a bit disconnected, since they're highly related and (IMO) should be able to be changed at the same time. I wonder what it would look like if `tools` were also stored on the chat models rather than LLMChain, and configured there instead (like @avergin's PR for tool_choice, but also for tools). It feels like they should either be both stored in LLMChain, or both stored in the chat model 🤔 

I had a look at how langchain.js works in this regard, and it looks like the tools and tool choice are stored in the chat model instance (working with shared interfaces and model-specific interfaces), but they can be changed [at the same time](https://github.com/langchain-ai/langchainjs/blob/b311ec5c19cd4ab7aad116e81fb1ea33c5d71a8d/libs/langchain-openai/src/chat_models.ts#L530-L538) with `llm.bindTools([tool], { tool_choice: ...} )`.

Thanks for having the discussion @brainlid! I don't know what the best solution is here, but I'll be happy with whatever you decide to go with 😄

PR #142, opened 2024-06-21
Add tool_choice for OpenAI and Anthropic

avergin:
This PR adds `tool_choice` option to both OpenAI and Anthropic models.

For both models, I've created a `tool_choice` attribute inside the model structs that embeds a `ToolChoice` struct with attributes that are specific to each model.

For `ChatOpenAI`:
```elixir
  defmodule ToolChoice do
    use Ecto.Schema

    defmodule Function do
      use Ecto.Schema
      @primary_key false
      embedded_schema do
        field :name, :string
      end

      @doc """
      Setup a ChatOpenAI Function with name
      """
      @spec changeset(%Function{}, map()) :: Ecto.Changeset.t()
      def changeset(%Function{}=function, %{}=attrs) do
        function
        |> cast(attrs, [:name])
        |> validate_required([:name])
      end

    end

    @primary_key false
    embedded_schema do
      field :type, Ecto.Enum, values: [:function, :required, :none, :auto]
      embeds_one :function, Function
    end

    @doc """
    Setup a ChatOpenAI ToolChoice configuration
    """
    @spec changeset(%ToolChoice{}, map()) :: Ecto.Changeset.t()
    def changeset(%ToolChoice{}=tool_choice, %{}=attrs) do
      tool_choice
      |> cast(attrs, [:type])
      |> validate_required([:type])
      |> validate_type()
    end

    defp validate_type(changeset) do
      case get_change(changeset, :type) do
        :function -> cast_embed(changeset, :function, required: true, with: &Function.changeset/2)
        _ -> changeset
      end
    end
  end
```

For `ChatAnthropic`:
```elixir
  defmodule ToolChoice do
    use Ecto.Schema

    @primary_key false
    embedded_schema do
      field :type, Ecto.Enum, values: [:tool, :auto, :any]
      field :name, :string
    end

    @doc """
    Setup a ChatAnthropic ToolChoice configuration
    """
    @spec changeset(%ToolChoice{}, map()) :: Ecto.Changeset.t()
    def changeset(%ToolChoice{}=tool_choice, %{}=attrs) do
      tool_choice
      |> cast(attrs, [:type, :name])
      |> validate_required([:type])
      |> validate_type()
    end

    defp validate_type(changeset) do
      case get_change(changeset, :type) do
        :tool -> validate_required(changeset, [:name])
        _ -> changeset
      end
    end
  end
```


brainlid:
Thank you for the contribution! In this particular case, I wonder if a struct is unnecessary? Especially since it will be the different for each model that supports it. 

Take a look at:

The recently added ChatOpenAI `stream_options`. It's just a very simple map. The validity is enforced by the LLM. This is also well supported `ChatOpenAI.serialize_config` function.

```elixir
field :stream_options, :map, default: nil
```

What do you think?

avergin:
I agree with you as this change will provide more flexibility to the library. I have replaced the structs with maps in the latest commit.

One caveat with this is: when setting `tool_choice` to `none`, `required` or `auto` in OpenAI, this will require the user to assign a map to the `ChatOpenAI` struct (e.g. `tool_choice: %{"type" => "required`}`). And it will raise an error when trying to set `tool_choice` directly with a string, rather than a map (e.g. `tool_choice: "required"`).

We might address this issue by defining a custom Ecto type, allowing tool_choice to accept either a string or a map. However, I'm not sure if this approach would be overkill, so I have not implemented it yet.

lib/chat_models/chat_anthropic.ex
```diff
@@ -201,11 +205,21 @@ defmodule LangChain.ChatModels.ChatAnthropic do
...
     |> Utils.conditionally_add_to_map(:top_k, anthropic.top_k)
   end
 
+  defp set_tool_choice(%ChatAnthropic{tool_choice: %{"type" => "tool", "name" => name}=_tool_choice}) when is_binary(name) and byte_size(name) > 0,
+    do: %{"type" => "tool", "name" => name}
+
+  defp set_tool_choice(%ChatAnthropic{tool_choice: %{"type" => type}=_tool_choice}) when type != nil and is_binary(type),
+    do: %{"type" => type}
+
+  defp set_tool_choice(%ChatAnthropic{}), do: nil
```
stevehodgkiss, line 220: Maybe `get_tool_choice` would be a better name?

stevehodgkiss (commented):
Looks good! It'd be worth adding a live test for tool choice. When implementing this in https://github.com/brainlid/langchain/pull/141 I had to modify these lines to get it to work: https://github.com/brainlid/langchain/pull/141/files#diff-98396b383d660bb5274db90a5fd13b1526d5d6b120a2d8ece0e8baf06e55b85eL695-R713

lib/chat_models/chat_anthropic.ex
avergin: Agree with you that it's a better name. Modified it accordingly.

Also added the tests for live calls.

test/chat_models/chat_open_ai_test.exs
```diff
@@ -689,6 +713,22 @@ defmodule LangChain.ChatModels.ChatOpenAITest do
...
+
+      {:ok, [message]} = ChatOpenAI.call(chat, [message], [weather, hello_world])
+
+      assert %Message{role: :assistant} = message
+      assert message.status == :complete
+      assert message.role == :assistant
+      assert message.content != nil
+      assert message.tool_calls == []
+    end
+
```
stevehodgkiss, line 731: Adding another live call test that forces the use of a tool triggers the bug I mentioned:

```elixir
    @tag live_call: true, live_open_ai: true
    test "executing a call with required tool_choice", %{
      weather: weather,
      hello_world: hello_world
    } do
      {:ok, chat} =
        ChatOpenAI.new(%{
          seed: 0,
          stream: false,
          model: @gpt4,
          tool_choice: %{"type" => "function", "function" => %{"name" => "get_weather"}}
        })

      {:ok, message} =
        Message.new_user("What is the weather like in Moab Utah?")

      {:ok, [message]} = ChatOpenAI.call(chat, [message], [weather, hello_world])

      # assertions...
    end
```

```
  1) test call/2 executing a call with required tool_choice (LangChain.ChatModels.ChatOpenAITest)
     test/chat_models/chat_open_ai_test.exs:737
     ** (WithClauseError) no with clause matching: %{"function" => %{"arguments" => "{\"city\":\"Moab\",\"state\":\"UT\"}", "name" => "get_weather"}, "id" => "call_q5WG1gd5X8gCqf9UlDceB3WA", "type" => "function"}
     code: {:ok, [message]} = ChatOpenAI.call(chat, [message], [weather, hello_world])
     stacktrace:
       (langchain 0.3.0-rc.0) lib/message.ex:225: anonymous fn/1 in LangChain.Message.validate_and_parse_tool_calls/1
       (elixir 1.16.2) lib/enum.ex:1700: Enum."-map/2-lists^map/1-1-"/2
       (langchain 0.3.0-rc.0) lib/message.ex:224: LangChain.Message.validate_and_parse_tool_calls/1
       (langchain 0.3.0-rc.0) lib/message.ex:151: LangChain.Message.common_validations/1
       (langchain 0.3.0-rc.0) lib/message.ex:121: LangChain.Message.new/1
       (langchain 0.3.0-rc.0) lib/chat_models/chat_open_ai.ex:825: LangChain.ChatModels.ChatOpenAI.do_process_response/2
       (elixir 1.16.2) lib/enum.ex:1700: Enum."-map/2-lists^map/1-1-"/2
       (langchain 0.3.0-rc.0) lib/chat_models/chat_open_ai.ex:552: LangChain.ChatModels.ChatOpenAI.do_api_request/4
       (langchain 0.3.0-rc.0) lib/chat_models/chat_open_ai.ex:473: LangChain.ChatModels.ChatOpenAI.call/3
       test/chat_models/chat_open_ai_test.exs:752: (test)
```

These changes fix that issue: https://github.com/brainlid/langchain/pull/141/files#diff-98396b383d660bb5274db90a5fd13b1526d5d6b120a2d8ece0e8baf06e55b85eL695-R713

avergin: This [commit](https://github.com/brainlid/langchain/pull/142/commits/d298843434342edf59369a8394e62bc7db875b9e) includes the fix as you described, as well as the live call test that produced the bug previously.

brainlid (approved):
Thanks for the contribution! It's a great addition to have.

❤️💛💙💜

test/chat_models/chat_open_ai_test.exs
brainlid: Thanks @stevehodgkiss for the help reviewing!

PR #156, opened 2024-07-18
Fix OpenAI chat stream hanging

stevehodgkiss:
We recently noticed that chat streams would appear to "hang" in the middle of generating a response. After a bit of debugging I found that it was due to the message being split in the middle of the "data: " part (i.e. "dat" in the last message, then "a: {" in the next). This results in the stream failing to parse any future messages and appearing to "hang" or freeze.

Haven't seen this issue before so I'm guessing something has changed in the way OpenAI streams their responses.

I couldn't see any tests for this particular case in the official node & python clients, but I modified [one of the node clients tests](https://github.com/openai/openai-node/blob/master/tests/streaming.test.ts#L37-L38) to check and it is supported (still passes). The node client appears to [split by line endings](https://github.com/openai/openai-node/blob/5873a017f0f2040ef97040a8df19c5b4dc2a66fd/src/streaming.ts#L383) first, then by `:` to [get the field name](https://github.com/openai/openai-node/blob/5873a017f0f2040ef97040a8df19c5b4dc2a66fd/src/streaming.ts#L330) (i.e. `data`).

The change in this PR pattern matches on their being something in the buffer (previous/partial message data), combines the buffered & new data then recursively calls the `decode_stream` function, since the combined data may need splitting again on the "data: " key.

### DIFF
--- a/lib/chat_models/chat_open_ai.ex
+++ b/lib/chat_models/chat_open_ai.ex
@@ -594,7 +594,7 @@ defmodule LangChain.ChatModels.ChatOpenAI do
   buffer data from a previous call, and assembling it to parse.
   """
   @spec decode_stream({String.t(), String.t()}) :: {%{String.t() => any()}}
-  def decode_stream({raw_data, buffer}) do
+  def decode_stream({raw_data, buffer}, done \\ []) do
     # Data comes back like this:
     #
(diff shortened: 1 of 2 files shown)


brainlid (approved):
Awesome @stevehodgkiss! Thanks! I just noticed that hanging behavior  the other day but hadn't diagnosed it yet.

❤️💛💙💜

PR #157, opened 2024-07-18
Remove ApiOverride reference in mix.exs project.docs

stevehodgkiss:
Removing a leftover reference to `ApiOverride` after the switch to Mimic in https://github.com/brainlid/langchain/pull/155


brainlid (approved):
Thanks!

❤️💛💙💜

PR #161, opened 2024-07-22
Update Azure OpenAI Service fix

stevehodgkiss:
Re: https://github.com/brainlid/langchain/pull/158

Changes the approach to handle this case in ChatOpenAI rather than more generally in `Utils.handle_stream_fn`.


brainlid (approved):
Thank you! Looks great.

❤️💛💙💜

PR #167, opened 2024-08-17
Fix PromptTemplate example

joelpaulkoch:
This gave me an error because it doesn't match `%LangChain.Message{}`


brainlid (approved):
Thanks for the doc improvement!

❤️💛💙💜

PR #171, opened 2024-08-25
Make api keys optional when running tests.

matthusby:
When running the test without having all the API keys set this would blow up even if you are not testing that api.  This will just try to load them and continue.


brainlid:
Hi @matthusby! I see where you're coming from with this. At the same time, having it error loudly has helped me when I moved development to another machine and the ENV wasn't all the way setup yet.

I'll think a bit more about this.

matthusby:
Yeah I can see that being useful.  I was coming at this from the default should just work, but if you are testing those api's then this does in fact break the just work part.

How do you feel about logging a notice if the key could not be found?

brainlid:
I'm not going to go in this direction at this time. Thanks for the thoughtful contribution!

PR #173, opened 2024-08-28
Structured Outputs by supplying `strict: true`

cristineguadelupe:
https://platform.openai.com/docs/guides/function-calling

> When you enable Structured Outputs by supplying strict: true, the OpenAI API will pre-process your supplied schema on your first request, and then use this artifact to constrain the model to your schema.


brainlid:
Thanks @cristineguadelupe! I'd seen the change from OpenAI but I'm super happy you added it!

❤️💛💙💜

PR #174, opened 2024-08-29
Do not duplicate tool call parameters if they are identical

michalwarda:
When using streaming with tools in ChatGoogleAI I've noticed that the merge method in tool_calls duplicated the name of the function. 

Because of that the tool calls were not working. For some reason ChatGoogleAI sends the same message delta with tools twice.

Below you can see that the name of the function was being duplicated: `image_1__generate_image` converted into `image_1__generate_imageimage_1__generate_image`.

```
primary: %LangChain.MessageDelta{
  content: nil,
  status: :complete,
  index: 0,
  role: :assistant,
  tool_calls: [
    %LangChain.Message.ToolCall{
      status: :incomplete,
      type: :function,
      call_id: "call-image_1__generate_image",
      name: "image_1__generate_image",
      arguments: %{"prompt" => "a dog"},
      index: nil
    }
  ]
}
delta call: %LangChain.Message.ToolCall{
  status: :incomplete,
  type: :function,
  call_id: "call-image_1__generate_image",
  name: "image_1__generate_image",
  arguments: %{"prompt" => "a dog"},
  index: nil
}
merged call: %LangChain.Message.ToolCall{
  status: :incomplete,
  type: :function,
  call_id: "call-image_1__generate_image",
  name: "image_1__generate_imageimage_1__generate_image",
  arguments: %{"prompt" => "a dog"},
  index: nil
}
```

I've introduced a fix that makes same messagedelta with tool call not duplicate

### DIFF
--- a/lib/message/tool_call.ex
+++ b/lib/message/tool_call.ex
@@ -152,6 +152,11 @@ defmodule LangChain.Message.ToolCall do
     |> update_status(call_part)
   end
 
+  defp append_tool_name(%ToolCall{name: primary_name} = primary, %ToolCall{name: new_name})
+       when primary_name == new_name do
+    primary
+  end
+
   defp append_tool_name(%ToolCall{} = primary, %ToolCall{name: new_name})
        when is_binary(new_name) do
     %ToolCall{primary | name: (primary.name || "") <> new_name}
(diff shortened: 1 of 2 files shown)


brainlid:
Yeesh! Thanks @michalwarda for figuring that out.

❤️💛💙💜

PR #175, opened 2024-09-05
🐛 cast tool_calls arguments correctly inside message_deltas

rparcus:
Before this fix, tool call arguments in deltas which contained several empty spaces were missing data when being cast.
While triaging this issue I also added some extra tests checking for correct casting of spaces.

### DIFF
--- a/test/chat_models/chat_open_ai_test.exs
+++ b/test/chat_models/chat_open_ai_test.exs
@@ -1233,6 +1233,31 @@ defmodule LangChain.ChatModels.ChatOpenAITest do
       assert parsed == [json_1, json_2]
     end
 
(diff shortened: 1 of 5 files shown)


test/chat_models/chat_open_ai_test.exs
```diff
@@ -1233,6 +1233,31 @@ defmodule LangChain.ChatModels.ChatOpenAITest do
       assert parsed == [json_1, json_2]
     end
 
+    test "correctly parses when data content contains spaces such as python code with intedentation" do
```
rparcus, line 1236:
```suggestion
    test "correctly parses when data content contains spaces such as python code with indentation" do
```

brainlid:
Great catch! Thanks for the improvement and additional tests!

❤️💛💙💜

PR #164, opened 2024-07-26
Enable tool support for ollama

alappe:
Adds basic support for tools with ollama `v0.3`, tested with the hairbrush example:

```elixir
alias LangChain.Function
alias LangChain.Message
alias LangChain.Chains.LLMChain
alias LangChain.ChatModels.ChatOllamaAI


# map of data we want to be passed as `context` to the function when
# executed.
custom_context = %{
  "user_id" => 123,
  "hairbrush" => "drawer",
  "dog" => "backyard",
  "sandwich" => "kitchen"
}

# a custom Elixir function made available to the LLM
custom_fn =
  Function.new!(%{
    name: "custom",
    description: "Returns the location of the requested element or item.",
    parameters_schema: %{
      type: "object",
      properties: %{
        thing: %{
          type: "string",
          description: "The thing whose location is being requested."
        }
      },
      required: ["thing"]
    },
    function: fn %{"thing" => thing} = _arguments, context ->
      # our context is a pretend item/location location map
      {:ok, context[thing]}
  end
  })

# create and run the chain
{:ok, updated_chain, %Message{} = message} =
  LLMChain.new!(%{
    llm: ChatOllamaAI.new!(%{model: "llama3.1", verbose: true}),
    custom_context: custom_context,
    verbose: true
  })
  |> LLMChain.add_tools(custom_fn)
  |> LLMChain.add_message(Message.new_user!("Where is the hairbrush located?"))
  |> LLMChain.run(mode: :while_needs_response)

# print the LLM's answer
IO.puts(message.content)
#=> "The hairbrush is located in the drawer."
```

PR #179, opened 2024-09-17
Fix options being passed to the ollama chat api

alappe:
I noticed that https://github.com/ollama/ollama/blob/main/docs/api.md#generate-a-chat-completion and https://github.com/ollama/ollama/blob/main/docs/modelfile.md#valid-parameters-and-values document that those options all need to be passed to the `options: {}` sub structure.

My local instance testing with llama `v3.1` and ollama `v0.3.10` complains because `num_gqa` is unknown. I left it in there for now because other models might need it.

### DIFF
--- a/lib/chat_models/chat_ollama_ai.ex
+++ b/lib/chat_models/chat_ollama_ai.ex
@@ -215,26 +215,27 @@ defmodule LangChain.ChatModels.ChatOllamaAI do
   def for_api(%ChatOllamaAI{} = model, messages, tools) do
     %{
       model: model.model,
-      temperature: model.temperature,
       messages: messages_for_api(messages),
       stream: model.stream,
-      seed: model.seed,
-      num_ctx: model.num_ctx,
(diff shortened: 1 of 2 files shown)


brainlid:
Hi @alappe! I assume this is still a needed change, right? The doc links you provided sure make it appear that way. (Thanks for those, btw!)

@mustela, your recent note [here](https://github.com/brainlid/langchain/pull/164#issuecomment-2690986216) leads me to believe you're successfully using the library with Ollama. Are you doing it with or without this change?

I'm trying to understand if making this change breaks backward compatibility or if this is just how it must be done going forward. Thanks!

alappe:
Hi @brainlid,

yes, this is still needed. It's still possible to use the library with ollama at the moment, because if you're not providing the options (in the right place), the defaults apply. But you cannot change options without the change. IMHO it breaks nothing because the current state just didn't work right with ollama.

alappe:
@brainlid I rebased the PR…

PR #181, opened 2024-09-27
Handle empty text parts from GoogleAI responses

elliotb:
When using a newer model variant such as `gemini-1.5-pro-002`, we occasionally receive empty text parts in the response, which was causing the validation of `ContentPart`s to fail. 

This PR filters those before creating the content part.


brainlid (approved):
Looks good! Thanks!

❤️💛💙💜

PR #182, opened 2024-09-27
Support system instructions for Google AI

elliotb:
The Google AI API now supports system instructions - https://ai.google.dev/gemini-api/docs/system-instructions?lang=rest

I've extracted the split_system_message function from ChatAnthropic and then handled processing the system message in the same way.


brainlid (approved):
Nice! I appreciate that you found an existing function and refactored it for reuse.

❤️💛💙💜

PR #183, opened 2024-09-27
Handle functions with no parameters for Google AI

elliotb:
If a function is defined with an empty parameter list, the API returns `* GenerateContentRequest.tools[0].function_declarations[0].parameters.properties: should be non-empty for OBJECT type\n` with the current implementation.

Removing the parameters key from the function declaration when there are no parameters resolves this.


brainlid (approved):
Thanks!

❤️💛💙💜

PR #184, opened 2024-09-27
Handle missing token usage fields for Google AI

elliotb:
Very sporadically, we see errors due to the usageMetadata map missing one of the token count fields when using the Google AI API.

This PR provides a fallback value when fetching either of the fields.


brainlid (approved):
Excellent. Thanks!

❤️💛💙💜

PR #186, opened 2024-10-02
Add support for passing safety settings to Google AI

elliotb:
This implements a new field on ChatGoogleAI for setting the safety settings for content filtering, following the documentation at https://ai.google.dev/gemini-api/docs/safety-settings#web and https://ai.google.dev/api/generate-content#v1beta.SafetySetting

Based on your comment at https://github.com/brainlid/langchain/pull/142#issuecomment-2183694933 I kept this as a simple list of maps, rather than introducing a struct and associated changesets, on the basis that the LLM will validate the settings.

I've added support for serializing the config, which should be backwards compatible as deserializing maps without the safety settings field will construct a model with the safety settings set to an empty list, which then wouldn't be added to the request as the list is empty.

### DIFF
--- a/lib/chat_models/chat_google_ai.ex
+++ b/lib/chat_models/chat_google_ai.ex
@@ -100,7 +100,8 @@ defmodule LangChain.ChatModels.ChatGoogleAI do
     :top_k,
     :receive_timeout,
     :stream,
-    :callbacks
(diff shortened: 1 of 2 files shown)


brainlid:
Nice! Thanks!

❤️💛💙💜

PR #187, opened 2024-10-04
Remove unused assignment from ChatGoogleAI

elliotb:
This code iterated over parts in the response, but the result was never used, as text_content was captured above and used instead.


brainlid:
Good catch! Thanks!

❤️💛💙💜

Issue #177, opened 2024-09-11
Make {:error, %Req.TransportError{reason: :timeout}} retriable


rparcus:
Quick question:
Why do we make `:timeout` as a non retryable error response? 
Req considers it a transient error:

<img width="794" alt="image" src="https://github.com/user-attachments/assets/a2277eaa-2fa5-4dec-a25c-006ff8fb2f1c">

brainlid:
It could be made to auto-retry I think. Perhaps with a built-in delay. I'm open to a PR for this.

PR #194, opened 2024-11-22
Revamped error handling and handles Anthropic's "overload_error"

### LINKED ISSUE #159: Return more than just text reason for errors.
When using llmchain when there is an error while invoking the completion returned the error itself is only a text. This makes it hard to correctly react to those messages and makes it super flaky in case the message changes. ie. when calling ChatOpenAI with too many tokens the result of chain run is:

`{:error, _chain, "This model's maximum context length is 16385 tokens. However, your messages resulted in 34966 tokens. Please reduce the length of the messages."}`

To react to this specific error type and ie. roll the memory one would need to do some `String.contains?("maximum context length")` to check if this is the specific error that happened or not. If the error text changes in the future for whatever reason our implementation will start failing. OpenAI returns the type of the error as a separate `type` field with value `context_length_exceeded` which is much easier to interpret later in code. Currently there is no way to retrieve this value from the `reason` returned from run.

Suggested solution. 

We can change the API to return error as ie. struct with "message" and optionally "code" or w/e. This is obviously a breaking change.

Another one is to return stringified JSON as the reason ie. for ChatOpenAI and let the user decide on which parts of error they want to use to interpret the error. This is also a semi-breaking change. It does not break the contract but the behavior is changed.

The last one I can think of is to pass custom callbacks for each of the error types to handle but I feel this is a pretty universal non provider dependent problem and still for the sync non streamed calling it would be cool to have the whole error as part of the returned value.

### LINKED ISSUE #193: Handle Anthropic overloaded errors during streaming
As we're making more use of streaming we're seeing more of these errors recently:

```elixir
[error] Unsupported event received when parsing Anthropic response: "event: error\ndata: {\"type\":\"error\",\"error\":{\"details\":null,\"type\":\"overloaded_error\",\"message\":\"Overloaded\"}        }"
```

The result is this error logged but the code seems to proceed like everything is okay. I would expect/prefer an error were raised so we could retry the conversation versus leaving it in a partial and incomplete state.

I started spelunking today but it looks like handling this could be a decent sized changes so I thought it was best to bring it here for discussion.

Anthropic documentation around streaming errors for reference: https://docs.anthropic.com/en/api/messages-streaming#error-events

brainlid:
This is a major change to how errors are handled. Instead of `LLMChain.run` returning `{:error, updated_chain, string_error}`, it now returns `{:error, updated_chain, %LangChainError{}}`.

LangChainError has a `message`, which is the original string message being returned before. However, it now also has a `type` which is a string code like "invalid_json", "overload_error", "changeset", "unexpected_response", "timeout", "retries_exceeded", etc.

This makes it easier to detect an error type and take corrective action.

This is a big change that touches all the Chat models. I tested all I could. 

**What is incomplete:**
- I'm sure I missed some documentation updates
- Some Chat models either don't have live tests, or I'm not setup for testing them.

**What you need to do:**
- Check you application code for how it is responding to and handling error responses.

Impacting:
- Closes #193 
- Closes #159
- References #177 - makes it easier to detect and handle a retry in application code.

### DIFF
--- a/lib/chains/data_extraction_chain.ex
+++ b/lib/chains/data_extraction_chain.ex
@@ -73,6 +73,7 @@ defmodule LangChain.Chains.DataExtractionChain do
   alias LangChain.PromptTemplate
   alias LangChain.Message
   alias LangChain.Message.ToolCall
+  alias LangChain.LangChainError
   alias LangChain.Chains.LLMChain
   alias LangChain.ChatModels.ChatOpenAI
 
@@ -86,7 +87,7 @@ Passage:
   Run the data extraction chain.
   """
   @spec run(ChatOpenAI.t(), json_schema :: map(), prompt :: [any()], opts :: Keyword.t()) ::
-          {:ok, result :: [any()]} | {:error, String.t()}
+          {:ok, result :: [any()]} | {:error, LangChainError.t()}
   def run(llm, json_schema, prompt, opts \\ []) do
     verbose = Keyword.get(opts, :verbose, false)
 
@@ -123,7 +124,7 @@ Passage:
           {:ok, info}
 
         other ->
-          {:error, "Unexpected response. #{inspect(other)}"}
+          {:error, LangChainError.exception("Unexpected response. #{inspect(other)}")}
       end
     rescue
       exception ->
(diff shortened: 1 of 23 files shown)


brainlid:
@stevehodgkiss When you have a chance, can you check how this change works for Bedrock support?

brainlid:
I've test this with Bedrock and it seems good.

PR #197, opened 2024-11-28
fix: enable verbose_deltas

cristineguadelupe:


brainlid:
Thanks! Great catch!
❤️💛💙💜

PR #198, opened 2024-12-01
Add LLAMA 3.1 Json tool call with Bumblebee

marcnnn:
This PR adds the Basic JSON tool calling functionality for LLAMA 3.1

The output is not as stable as I hoped it would be, maybe I need to adjust the template a little.

The tool calling will only be used if you use the template:

```elixir
LangChain.ChatModels.ChatBumblebee.new!(%{
      serving: Llama,
      template_format: :llama_3_1_json_tool_calling,
      stream: false
    })
```


marcnnn:
In the example, I added:
`Do not mention that you used function calling to the user.`
To the System prompt with that, the output was stable.

@brainlid before merge in the example notebook, we should change that import, if you like everything else so far.

README.md
```diff
@@ -201,6 +201,7 @@ For example, if a locally running service provided that feature, the following c
 Bumblebee hosted chat models are supported. There is built-in support for Llama 2, Mistral, and Zephyr models.
 
 Currently, function calling is NOT supported with these models.
```
xhr15, line 203: Maybe rephrase this to 
"Currently, function calling is only supported for llama 3.1  Tool calling for Llama 2, Mistral, and Zephyr is NOT supported.

lib/chat_models/chat_bumblebee.ex
```diff
@@ -236,6 +237,16 @@ defmodule LangChain.ChatModels.ChatBumblebee do
   @doc false
   @spec do_serving_request(t(), [Message.t()], [Function.t()]) ::
           list() | struct() | {:error, String.t()}
+
+  def do_serving_request(%ChatBumblebee{template_format: :llama_3_1_json_tool_calling} = model, messages, functions) do
+    prompt = ChatTemplates.apply_chat_template_with_tools!(messages, model.template_format,functions)
+    |> IO.inspect
```
brainlid, line 243: Remove the IO.inspect

brainlid (changes requested):
It looks like the code formatter could also be run.

Thanks for all this work!

brainlid:
> @brainlid before merge in the example notebook, we should change that import, if you like everything else so far.

Sorry, change what import?

notebooks/bumblebee_llama3_1_json_tooluse.livemd
```diff
@@ -0,0 +1,98 @@
...
+
+```elixir
+Mix.install(
+  [
+    {:kino_bumblebee, ">= 0.5.0"},
+    {:bumblebee, ">= 0.5.3"},
+    {:exla, ">= 0.0.0"},
+    {:axon, ">= 0.5.1"},
+    {:nx, ">= 0.5.1"},
+    {:langchain, github: "marcnnn/langchain", branch: "llama3_1_json_tool"}
```
marcnnn, line 11: This one should import just the latest version after published.

marcnnn:
Thanks for the feedback!
Since you saw no big problems with the implementation,
I will continue to look into other tool calling methods and other models.

lib/chat_models/chat_bumblebee.ex
```diff
@@ -244,7 +288,148 @@ defmodule LangChain.ChatModels.ChatBumblebee do
...
+    end
+  end
+
+  def do_process_response(
+        %{results: [%{text: "<" <> _ = content, token_summary: token_summary}]},
+        %ChatBumblebee{template_format: :llama_3_1_custom_tool_calling} = model
+      )
+      when is_binary(content) do
+    fire_token_usage_callback(model, token_summary)
+    IO.inspect(content)
```
brainlid, line 343: Remove the IO.inspect?

lib/chat_models/chat_bumblebee.ex
```diff
@@ -123,7 +126,8 @@ defmodule LangChain.ChatModels.ChatBumblebee do
     # # more focused and deterministic.
     # field :temperature, :float, default: 1.0
 
-    field :template_format, Ecto.Enum, values: [:inst, :im_start, :zephyr, :llama_2, :llama_3]
+    field :template_format, Ecto.Enum,
+      values: [:inst, :im_start, :zephyr, :llama_2, :llama_3, :llama_3_1_json_tool_calling, :llama_3_1_custom_tool_calling, :llama_3_2_custom_tool_calling]
```
brainlid, line 130: This is the line where I think the merge conflict is. It should be a very simple fix though.

lib/utils/chat_templates.ex
```diff
@@ -314,11 +317,319 @@ defmodule LangChain.Utils.ChatTemplates do
...
+end
+
+def llama_3_1_custom_tool_calling_parameter_conversion(tools) do
+  tools
+  |> Enum.map(fn %LangChain.Function{name: name, description: description, parameters_schema: schema} ->
+    props = schema[:properties]
+    parameters =
+      props
+      |> Enum.map(fn {param_name, param_config} ->
+        IO.inspect(param_config)
```
brainlid, line 597: IO.inspect to remove

lib/utils/chat_templates.ex
```diff
@@ -314,11 +317,319 @@ defmodule LangChain.Utils.ChatTemplates do
...
+
+  # Format tools as simple JSON schema
+  tools_json_schema_string =
+    llama_3_1_custom_tool_calling_parameter_conversion(tools)
+
+  text = """
+  <|begin_of_text|>
+  <|start_header_id|>system<|end_header_id|>
+
+  Environment: ipython
```
brainlid, line 463: The `ipython` bit surprises me a bit! I'm sure it makes sense somehow. :)

lib/utils/parser/llama3_2_custom_tool_paser.ex
```diff
@@ -0,0 +1,78 @@
+defmodule Langchain.Utils.Parser.LLAMA_3_2_CustomToolParser do
+  import NimbleParsec
```
brainlid, line 2: I noticed this PR adds the nimble_parsec dependency and it uses the "defparsec" macro.

Can you explain a bit more about what it's help to solve? I'm careful to not add dependencies that aren't needed, or it could be solved in a way that didn't require the dep.

notebooks/bumblebee_llama3_1_json_tooluse.livemd
brainlid: Ah, yes. I get what you mean.

brainlid (commented):
This is a very meaty change! That's some impressive work. 

My biggest question is about nimble_parsec and needing to resolve the merge conflict. But I think it looks good. 🙂

lib/utils/parser/llama3_2_custom_tool_paser.ex
marcnnn: The Tool call output of Llama3.2 is not just JSON.
It is a python like list of function calls.

Using parsec makes reading, understanding and changing much easier than self written parser code.
Since many Security Problems in Programming are in parser Code, I decided to use a parser generator when ever possible.
And I am used to writing context free grammar, it is the natural why to solve parsing.

If an argument can be a key value Map / dictionary / (JSON/Python) Object or a List, then I would think it's harder to understand without a parser generator.
But I have not found documentation on Llama3.2 support for that.
If someone is interested, let my know or test it out and let me know.

@brainlid Is there an easy way to parse that without dependencies?

brainlid: @marcnnn,

Perhaps the best option is to make the dependency optional and check that it's available in the Bumblebee-related code. It could even error if not present and tell them the dependency is needed.

Like this: 
```elixir
{:nimble_parsec, "~> 1.4", optional: true}
```

Then when using it:

```elixir
  if Code.ensure_loaded?(NimbleParsec) do
    # ... use the code ...
  else
    # raise error that optional dependency is required
  end
```

Also, is there a specific LLaMa version that is required for this to work? Is it a fine-tuned derivative model? Is is the stock model available from Meta? 

The docs should probably include a link to a version of the model that this is designed to work with.

lib/utils/chat_templates.ex
marcnnn: https://www.llama.com/docs/model-cards-and-prompt-formats/llama3_1
Under the topic:

User-defined Custom tool calling

This is what Meta trained with.

lib/utils/parser/llama3_2_custom_tool_paser.ex
marcnnn: It works with Meta Llama 3.[1-3] instruction-turned models:

https://www.llama.com/docs/model-cards-and-prompt-formats/llama3_1

https://www.llama.com/docs/model-cards-and-prompt-formats/llama3_2

https://www.llama.com/docs/model-cards-and-prompt-formats/llama3_3

I can do that and check only on the templates that need parsec...

I would do the `Code.ensure_loaded?(NimbleParsec)` before the LLM is called for the User to fail fast, even if this means that the code is not close together?

Llama 3.3 was not tested since it is too big for one GPU and distribution is currently worked on by Paulo.

I could add my Livebooks as examples in a separate PR to not add too much at once?

brainlid: Thanks for the links! We can include those in the docs as a resource.

As for the `Code.ensure_loaded?` check, I lean towards keeping it in the custom tool parser usage. The reason is, I might validly use Llama 3 without any intention of using tool calls. 

Also, hitting the issue will happen during development time as the developer tries to use that feature. That feels like a fine time and developer experience to me. 

> Llama 3.3 was not tested since it is too big for one GPU and distribution is currently worked on by Paulo.

That makes sense. Don't worry about that. :slightly_smiling_face:

marcnnn: Parsec is Optional now and is only needed for custom tool calling.

brainlid:
@marcnnn, I'm not sure if this is ready yet. Just let me know when you're happy with where it is. :slightly_smiling_face:

marcnnn:
I would say I am ready

brainlid:
Awesome @marcnnn!

Thanks for working through all that and bringing this great enhancement.
❤️💛💙💜

PR #222, opened 2025-01-03
Remove constraint of alternating user, assistant

GenericJam:
I'm doing something a bit funky and I need to be able to have messages in whatever order. 
eg system/user/assistant/user1/user2/assistant/user2/user1/assistant/etc.
The LLM (Ollama) doesn't seem to mind.

### DIFF
--- a/lib/utils/chat_templates.ex
+++ b/lib/utils/chat_templates.ex
@@ -104,8 +104,11 @@ defmodule LangChain.Utils.ChatTemplates do
   If there is an issue, an exception is raised. Reasons for an exception:
 
   - Only 1 system message is allowed and, if included, it is the first message.
-  - Non-system messages must begin with a user message
-  - Alternates message roles between: user, assistant, user, assistant, etc.
+  - Non-system messages must begin with a user message or assitant message.
+
+  Recent change:
+  - Alternating messages between user / assistant / user / assistant are no longer enforced as not every model has issues.
+  - It is up to the programmer to enforce this if this is something they need.
   """
   @spec prep_and_validate_messages([Message.t()]) ::
           {Message.t() | nil, Message.t(), [Message.t()]} | no_return()
@@ -138,17 +141,16 @@ defmodule LangChain.Utils.ChatTemplates do
     # must alternate user, assistant, user, etc. Put the first user message back
     # on the list for checking it.
     [first_user | rest]
-    |> Enum.with_index()
     |> Enum.each(fn
-      {%Message{role: :user}, index} when is_even(index) ->
+      %Message{role: :user} ->
         :ok
 
-      {%Message{role: :assistant}, index} when is_odd(index) ->
+      %Message{role: :assistant} ->
         :ok
 
       _other ->
         raise LangChainError,
-              "Conversation roles must alternate user/assistant/user/assistant/..."
+              "Conversation roles must be either user or assistant."
     end)
 
     # return 3 element tuple of critical message pieces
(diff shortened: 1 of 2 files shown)


jeroenbourgois:
@GenericJam is it really necessary to have this in the codebase as an option? You can just as easily sort the result in whatever order you need? I am **not** being hostile, but over the years I have seen many libraries that offer too many config options (which always leads to more complex code) with some of those options being used very sparsely.

GenericJam:
@jeroenbourgois I agree about too many options making the code brittle, annoying, etc.

The alternative for me if I want to use this library is to catch the exception, kill it and run the chain anyway which seems like even more of an antipattern.

I could also try to bunch multiple user messages into a single message but I think this will be more confusing for everyone, including the bot so I think that's also less desirable. 

Where I'm going with this. Here's an example system prompt:
You are a helpful assistant. There are several participants in this conversation. They are labelled before their text. You can respond to a participant by using @ before their name like @Lucy.
Example conversation:
user: "Lucy: What is 2+2?"
user: "Dave: What about 1+3?"
assistant: "@Lucy 2+2 is 4. @Dave 1+3 is also 4. Do you have any other questions?"

The goal here is to have multi bot and or multi human conversations. From a particular bot's perspective they are always placed as the assistant so I'm not actually reusing the chain, I rebuild it every time. What I am going for here is to be able to talk to more than one bot to get a second answer if one is not satisfactory.

The two factors that seem to make up a bot so far are its base model and the system prompt. I want to be able to make various bot personas this way. The user constructs them ahead of time and interacts with them in the conversation. This way a topic can be viewed from more than one perspective.

Are there any LLMs that actually have a problem if they receive non alternating conversations? In my testing so far I haven't encountered a problem.

GenericJam:
I understand that an option is annoying but I'm just trying to support base behaviour if there are people that actually want that. 
I don't know why it's an error at all. 
The LLMs don't choke on it so it's not like we're protecting from a crash. It seems excessive to crash for colouring outside the lines.

GenericJam:
Should be good now.

brainlid:
Thanks for the updates.

For context: we had an external conversation and I realized that the previously enforced rules defined in ChatTemplates on HuggingFace have been updated to no longer require the exact user -> assistant -> user flow.

Some models allow for it and others do not. It is up to the developer to know what their specific model requires.

brainlid:
> @jeroenbourgois 
> ...
> Where I'm going with this. Here's an example system prompt:
> 
> ```
> You are a helpful assistant. There are several participants in this conversation. They are labelled before their text. You can respond to a participant by using @ before their name like @Lucy.
> Example conversation:
> user: "Lucy: What is 2+2?"
> user: "Dave: What about 1+3?"
> assistant: "@Lucy 2+2 is 4. @Dave 1+3 is also 4. Do you have any other questions?"
> ```

First, I agree on updating to remove the constraint. Thanks for the changes!

I do not recommend this message approach though. I believe you will get better performance from something like this:

```
USER:
Recently received messages:
- Lucy: What is 2+2?
- Dave: What about 1+3?
ASSISTANT:
@Lucy 2+2 is 4. @Dave 1+3 is also 4. Do you have any other questions?
```

Or embedding the messages into an XML structure like:
```
USER:
Recently received messages:
<user name="Lucy">
What is 2+2?
</user>
<user name="Dave">
Dave: What about 1+3?
</user>
```

The approach would really depend on the capabilities of the model. But I believe something like this will work better in general.

PR #226, opened 2025-01-07
Add prompt caching support for Claude.

montebrown:
- both ChatGPT and Claude offer prefix-based prompt caching, however, unlike ChatGPT, prompt caching for Claude is not automatic. 
 - Claude will cache tokens up until a block marked with 'cache_control'. This can include tools, system, and messages, in that order.  
 - setting cache_control can now be done at the ContentPart level by setting the :cache_control option to true or %{type: "ephemeral"}. See ChatAnthropicTest for an example.

Notes:
 - this change means that system messages can contain a list of ContentParts now, so that cache_control can be set in the middle of a system message. 
 - this does not address collecting the usage metrics - when caching is used, cache_creation_input_tokens and cache_read_input_tokens are returned in the usage metadata in addition to the usual input_tokens and output_tokens.


brainlid:
Thanks @montebrown! 

I appreciate the update to the docs and tests. Great enhancement!
❤️💛💙💜

brainlid:
@montebrown, 

Heads up that this PR https://github.com/brainlid/langchain/pull/236 would expose the additional custom token usage information for cached parts.

PR #227, opened 2025-01-08
feat: Enable :inet6 for Req.new for Ollama

mpope9:
Tried to use this with fly.io. I was unable to connection to Ollama hosted on a private flycast address. Enabling ipv6 fixed the issue.


mpope9:
Also a suggestion: a deployable Ollama + Elixir example running on fly as it was a bit of a pain to figure out why it wouldn't connect 😅

brainlid:
Thanks for the IPv6 update! I don't have access to Fly.io GPU hardware anymore so I'm not in a position to create that sample project or more detailed guide. If you're offering, that would be cool!

mpope9:
Thanks for the merge! I'll see if I can come up with something sensible.

PR #231, opened 2025-01-09
feat: apply chat template from callback

joelpaulkoch:
I want to work with [Phi-4](https://huggingface.co/microsoft/phi-4) which has a slightly different chat template than previous versions.
I could add the template but I figured some time ago it would be practical to be able to pass a callback to `apply_chat_template!/3`.
This way, users don't have to wait for an release to use a new template.

So, here is a PR.
I might go ahead and add the `phi-4` template in a separate PR.

### DIFF
--- a/lib/utils/chat_templates.ex
+++ b/lib/utils/chat_templates.ex
@@ -79,11 +79,16 @@ defmodule LangChain.Utils.ChatTemplates do
 
   Note: The `:zephyr` format supports specific system messages.
 
+  ## Template callback
+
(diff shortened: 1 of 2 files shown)


brainlid:
Hi @joelpaulkoch! I love this approach. It provides a nice future-proof way of adding one-off support for some Bumblebee hosted model.

As an FYI, this was just merged: https://github.com/brainlid/langchain/pull/233

But I still want merge in your approach too.

PR #233, opened 2025-01-12
Add Bumblebee Phi-4

marcnnn:
This commit adds the Prompt template for the Phi-4 model.

Tested with quantized Phi-4 on 40GB A100. (17.7 GB vram and 100GB ram to quantize)

template_format: :phi_4

If I should add the livebook let me know?


brainlid:
Awesome! Thanks!
❤️💛💙💜

Issue #235, opened 2025-01-15
Extend TokenUsage to Support Custom Fields for Enhanced Usability


nallwhy:
Currently, the `TokenUsage` module in LangChain only supports `input` and `output` fields, limiting its flexibility for use cases requiring additional metadata, such as `cached_tokens`. By allowing extensions to include a `raw` field or other custom fields, users could store and retrieve more comprehensive information tailored to their specific needs.

```elixir
%LangChain.TokenUsage{
  input: 19,
  output: 10,
  raw: %{
    "completion_tokens" => 10,
    "completion_tokens_details" => %{
      "accepted_prediction_tokens" => 0,
      "reasoning_tokens" => 0,
      "rejected_prediction_tokens" => 0
    },
    "prompt_tokens" => 19,
    "prompt_tokens_details" => %{"cached_tokens" => 0},
    "total_tokens" => 29
  }
}
```

brainlid:
Closed by PR https://github.com/brainlid/langchain/pull/236

PR #236, opened 2025-01-16
Add raw field to TokenUsage

nallwhy:
This PR resovles #235.


brainlid:
This is a good solution to the problem of many different LLMs and their ever-changing set of usage statistics. Thanks!
❤️💛💙💜

PR #237, opened 2025-01-22
feat: Support for Ollama keep_alive API parameter

mpope9:
This API parameter should keep the model "warm" in memory. Otherwise after 5 minutes it is unloaded, and can be expensive to re-load.

Relevant docs: https://github.com/ollama/ollama/blob/main/docs/faq.md#how-do-i-keep-a-model-loaded-in-memory-or-make-it-unload-immediately


brainlid:
Nice! Thanks for the enhancement!
❤️💛💙💜

mpope9:
Thank you for this library!

PR #241, opened 2025-01-26
Include stacktrace context in messages for caught exceptions from LLM functions & function callbacks.

montebrown:
Hi! In upgrading some code to 0.3.0 I had a few exceptions to track down and found the lack of context in the caught error messages slowed things down.

This change adds context  to the error messages logged and kept in the chain for LLM functions and function callbacks. I took an initial stab at the places this seems most valuable while avoiding putting full stack traces in the chain. 

The details:
 - caught exceptions were logging their type and message, however, without a stack trace or location for the exception debugging was required to locate some issues. 
 - added format_exception/3 to LangChain.LangChainError to format exceptions with context from the stacktrace. The third argument is the format - :short only includes the location and the default gives a full stack trace at the end of the message.
 - I updated the most obvious places where a lack of context could slow down fixing issues, however there are other rescue blocks where this may be useful.

### DIFF
--- a/lib/callbacks.ex
+++ b/lib/callbacks.ex
@@ -42,7 +42,7 @@ defmodule LangChain.Callbacks do
           rescue
             err ->
               msg =
-                "Callback handler for #{inspect(callback_name)} raised an exception: #{inspect(err)}"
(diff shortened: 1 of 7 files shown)


brainlid:
Very nice! I love this. Thanks! I appreciate the updated test coverage too.
❤️💛💙💜

PR #249, opened 2025-02-06
typos: Update Example for Syntax Issues

bradschwartz:
Move the `to_string` call into the pipeline as it returns a Tuple which IO.puts can't handle:

```log
** (Protocol.UndefinedError) protocol String.Chars not implemented for {:ok, "The hairbrush is located in the drawer."} of type Tuple
```

Renamed the variable for consistency/compilation.


brainlid (approved):
Thanks for helping to improve the README docs!
❤️💛💙💜

PR #250, opened 2025-02-06
add native tool functionality (e.g. `google_search` for Gemini)

avergin:
Adds support for calling native tools like Google Search in Gemini models.

For example, `What is the current Google stock price?` will return the current stock price.

```elixir
alias LangChain.Message
alias LangChain.NativeTool

model = ChatGoogleAI.new!(%{temperature: 0, stream: false, model: "gemini-2.0-flash"})

{:ok, updated_chain} =
   %{llm: model, verbose: false, stream: false}
   |> LLMChain.new!()
   |> LLMChain.add_message(
     Message.new_user!("What is the current Google stock price?")
   )
   |> LLMChain.add_tools(NativeTool.new!(%{name: "google_search", configuration: %{}}))
   |> LLMChain.run()
```

I have also added a `metadata` field (map) to the `Message` struct to flexibly capture model specific responses when native tools are called. For example when `google_search` is called, Gemini returns `groundingMetadata` information, and this information can be accessed from `metadata`.

### DIFF
--- a/lib/chains/llm_chain.ex
+++ b/lib/chains/llm_chain.ex
@@ -133,6 +133,7 @@ defmodule LangChain.Chains.LLMChain do
   alias LangChain.Function
   alias LangChain.LangChainError
   alias LangChain.Utils
+  alias LangChain.NativeTool
 
   @primary_key false
   embedded_schema do
@@ -273,7 +274,7 @@ defmodule LangChain.Chains.LLMChain do
   @doc """
   Add a tool to an LLMChain.
   """
-  @spec add_tools(t(), Function.t() | [Function.t()]) :: t() | no_return()
+  @spec add_tools(t(), NativeTool.t() | Function.t() | [Function.t()]) :: t() | no_return()
   def add_tools(%LLMChain{tools: existing} = chain, tools) do
     updated = existing ++ List.wrap(tools)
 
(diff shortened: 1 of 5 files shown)


avergin:
For example, the `metadata` field in the `Message` struct may help with capturing the `citations` data that is returned from Perplexity's API response.

brainlid:
Hi @avergin! Thanks for doing the work to add this!

I like the NativeTool approach and yes, I can see it being useful for other models. The Message metadata is interesting too. Until things standardize on what types of results can be expected, then I think it makes sense. I'd prefer to have a more normalized result structure... but we don't have enough examples yet.

Can you please add some documentation to help people understand how to benefit from your work?
- Add to the `ChatGoogleAI` module doc a brief explanation of the supported feature and an example for how it can be used
- `NativeTool` a brief module doc that explains the purpose of the module. That it is for representing tools that are built-in and available from an AI/LLM service.
 
I think what you've added is interesting and valuable and I would like for it be more prominent and discoverable for people. I don't feel I can provide as good of an example usage perhaps as you can.

I'm happy to merge it in after the documentation is updated!

avergin:
Hi @brainlid! I'm glad that you found it valuable. I'm heavily using it in my project as well.

I've updated the documentation for both modules. Let me know if anything is missing!

brainlid (approved):
Awesome! Thanks!
❤️💛💙💜

PR #251, opened 2025-02-08
Introduce `Utils.TokenUsageCollector` for token usage tracking

nallwhy:
This PR introduces `LangChain.Utils.TokenUsageCollector`, a utility module designed to simplify token usage tracking.

**Motivation:**

Currently, while the `on_llm_token_usage` callback provides valuable token usage information, there isn't a readily available, best-practice approach for aggregating this data, especially in scenarios involving multiple LLM interactions within a single task. This PR aims to address this by providing a reusable and easy-to-use solution.

**Benefits:**

*   Simplified Token Tracking: Makes it easier for developers to track token usage across multiple LLM calls.
*   Inspiration for Callback Usage: Offers a practical example of how to effectively utilize the `on_llm_token_usage` callback.

I believe this module can be a valuable addition to `LangChain`, providing a more user-friendly experience for managing token usage. I'm open to feedback and suggestions for further improvements.

### DIFF
--- a/lib/utils/token_usage_collector.ex
+++ b/lib/utils/token_usage_collector.ex
@@ -1,0 +1,104 @@
+defmodule LangChain.Utils.TokenUsageCollector do
+  @moduledoc """
+  A utility module providing a token usage collector using an Agent.
+
+  This module offers a convenient and best-practice approach to collecting
+  token usage information within the LangChain library, specifically designed
+  for use with the `on_llm_token_usage` callback.
+
+  In scenarios where a single task involves multiple interactions with AI models,
+  this Agent-backed collector allows for easy aggregation of total token usage.
+
+  ## Example
+
+    iex> collector = LangChain.Utils.TokenUsageCollector.new()
+    iex> LangChain.Utils.TokenUsageCollector.acc(collector, %{
+    ...>   "gemini-2.0-flash" => %LangChain.TokenUsage{
+    ...>     input: 1,
+    ...>     output: 2,
+    ...>     raw: %{"promptTokenCount" => 1, "candidatesTokenCount" => 2, "totalTokenCount" => 3}
+    ...>   }
+    ...> })
+    iex> LangChain.Utils.TokenUsageCollector.acc(collector, %{
+    ...>   "gemini-2.0-flash" => %LangChain.TokenUsage{
+    ...>     input: 3,
+    ...>     output: 7,
+    ...>     raw: %{"promptTokenCount" => 3, "candidatesTokenCount" => 7, "totalTokenCount" => 10}
+    ...>   }
+    ...> })
+    iex> LangChain.Utils.TokenUsageCollector.acc(collector, %{
+    ...>   "gemini-1.5-pro" => %LangChain.TokenUsage{
+    ...>     input: 2,
+    ...>     output: 5,
+    ...>     raw: %{"promptTokenCount" => 2, "candidatesTokenCount" => 5, "totalTokenCount" => 7}
+    ...>   }
+    ...> })
+    iex> LangChain.Utils.TokenUsageCollector.get(collector)
+    %{
+      "gemini-2.0-flash" => %LangChain.TokenUsage{
+        input: 4,
+        output: 9,
+        raw: %{"promptTokenCount" => 4, "candidatesTokenCount" => 9, "totalTokenCount" => 13}
+      },
+      "gemini-1.5-pro" => %LangChain.TokenUsage{
+        input: 2,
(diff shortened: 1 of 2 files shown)


nallwhy:
@brainlid 
The `get_callback/1` function will likely error out when used with `ChatBumblebee` because `ChatBumblebee` doesn't have a `model` attribute in the callback data.  I'm looking for suggestions on how best to handle this case.

nallwhy:
@brainlid 

I find it challenging to implement the changes you suggested.

1. I’m not sure where exactly I should store the token usage within the process. Personally, I’m not very comfortable using the process dictionary or similar mechanisms, since if the process is reused, the dictionary values need to be explicitly cleared, leading to mutable and stateful behavior, which I’d prefer to avoid.
2. In my use cases, I often run multiple tasks concurrently using this library. Because of that, using a separate, independent process (such as an Agent) feels more natural and convenient to me. It keeps things isolated and works well with concurrent tasks.
3. Making token usage storage default to self, but also allowing it to be stored in other pids, seems to require some control over those external processes. However, I believe controlling external pids is generally not straightforward and could introduce complexity.

The approach I prefer is to start a new Agent when a large task begins, use it to accumulate token usage across all related tasks, and finally retrieve the total token usage at the end.

For example, something like this:
```elixir
token_usage_collector = TokenUsageCollector.new()

context = %{
  ...,
  token_usage_collector: token_usage_collector
}

result = large_task(inputs, context)

token_usage = TokenUsageCollector.get(collector)
```

aaparmeggiani:
What if `token_usage` were stored directly in each message (`message.token_usage` // `message.metadata.token_usage` ?  ), and  aggregated from the messages in `updated_chain` when needed?

```
message_usage = TokenUsage.get(%Message{} = message)
chain_usage = TokenUsage.get(updated_chain)
```

This approach would eliminate the need to manage state while keeping partials attached to messages for other potential use cases.

brainlid:
Thanks to your work here and on another PR, I think I've got a better solution for TokenUsage tracking. I'll let you know when I have it ready.

brainlid:
> What if `token_usage` were stored directly in each message (`message.token_usage` // `message.metadata.token_usage` ? ), and aggregated from the messages in `updated_chain` when needed?
> 
> ```
> message_usage = TokenUsage.get(%Message{} = message)
> chain_usage = TokenUsage.get(updated_chain)
> ```
> 
> This approach would eliminate the need to manage state while keeping partials attached to messages for other potential use cases.

I like the API for it too. :+1:

aaparmeggiani:
> The question of support really depends on when the LLM provides the data and how it's linked. In general I think it will work well.

Yeah, I'm not sure how different providers handle this, but your basic implementation covering `input`, `output`, and `total`  +  keeping  raw seems a good start (until people start mixing chains :-).

nallwhy:
@brainlid I also think integrating token usage directly into the message makes a lot of sense. Initially, I noticed it wasn’t implemented that way and instead relied on the callback approach for aggregating token usage. I assumed there must have been an intentional reason for that design choice, so I followed that direction in my implementation.

@aaparmeggiani  As shown in your example, aggregating usage per model seems like a solid approach. Even if the raw data differs due to model mixing, it shouldn’t be an issue. Since the input and output token costs vary between models (and the raw cache info can also affect pricing), I believe aggregating by model is generally the more accurate and practical solution.

aaparmeggiani:
I took a similar approach to @nallwhy, but in my case, not having access to the processed message in the callback was a dealbreaker. The example is just a PoC in need of refinement. Really glad our different use cases and feedback are helping improve this great library—thanks to both of you! 😊

manukall:
I also need to be able to track token usage per message for a project I'm working on. I'd love to help implement this. 

@brainlid would you be open to me working on this? If so, can you describe what you have in mind, so I don't start off in the wrong direction?

brainlid:
Hi @manukall! Thanks for offering, however, I'm going to include the TokenUsage change while I'm working on a bigger change for receiving streamed multi-modal responses from the AI. This is like the "thinking" content from Anthropic but could also be used for receiving images and text in a response.

manukall:
@brainlid Got it, thanks. I'm excited about what you'll come up with! :-)

brainlid:
TokenUsage is now being accumulated and stored on the received message in `%Message{metadata: %{usage: %TokenUsage{}}}` where available. 

As of v0.4.0-rc.0, this is implemented for ChatOpenAI and ChatAnthropic. I invite others to update other chat models in a similar way.

nallwhy:
@brainlid Cool!

Issue #252, opened 2025-02-10
`on_llm_token_usage` seems broken on master


fbettag:
i'm currently preparing a PR to add token usage and tool support to mistral. Sadly i've been chasing a ghost all morning since for some reason the callback for token usage does not get called anymore.

The regression seems to have been introduced sometime between current and `9e07845168281f85ea29db97e392d274a698feeb`  (November 2024).

I will launch the PR regardless shortly, but my apps somewhat depend on knowing how much user requests cost, so getting that working again would be good :)

fbettag:
ah the old `callbacks` paramter in .new/1 doesnt work anymore. `add_callback` seems to fix it! sorry

PR #253, opened 2025-02-10
adds tool calls and usage for mistral ai.

fbettag:
just really ported code over from ChatOpenAI as the implementations are similar.


fbettag:
ah damn, there are tests. wouldn't have guessed that from #252  :D will fix them too

fbettag:
i am using this as a drop in replacement for OpenAI on two projects and it works really really nice :)

brainlid:
Great! Thanks for your work on bringing up the Mistral AI support!

And thank you for adding the tests as well. :+1: 
❤️💛💙💜

PR #255, opened 2025-02-12
Enable structured outputs for Google AI

sglyon:
This quick change is something I needed locally to get structured JSON outputs from the gemini models: https://ai.google.dev/gemini-api/docs/structured-output?lang=rest

### DIFF
--- a/lib/chat_models/chat_google_ai.ex
+++ b/lib/chat_models/chat_google_ai.ex
@@ -84,6 +84,8 @@ defmodule LangChain.ChatModels.ChatGoogleAI do
 
     field :stream, :boolean, default: false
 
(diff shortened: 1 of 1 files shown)


brainlid:
Thanks for working to add structured output for Google Gemini.

I'm wondering, does adding structured output really remove the ability to set topP and topK?

I'd love to see some test coverage for the addition, to help ensure it doesn't have an unintended regression at some point.

And, assuming that the topP and topK are still valid, then the settings could be added to the config map in another way.

brainlid:
Thank you for your focus and contribution in this direction! Please refer to this merged PR as an alternate implementation that will hopefully meet your needs as well.

https://github.com/brainlid/langchain/pull/277

PR #257, opened 2025-02-13
ensure consisten capitalization

JoaquinIglesiasTurina:
Ensures consistent capitalization of library name.


brainlid:
Thanks for catching that and cleaning it up!
❤️💛💙💜

PR #260, opened 2025-02-20
Feature/support sys instruction vertexai

vseng:
I noticed that the data structure for vertexai's system instruction is different from some of the other chat models and that the current process of passing system instructions as messages with the :system role wasn't supported in the vertexai chat model.

This change allows the user to optionally pass the system instruction when creating a new vertexai chain.


brainlid:
Hi @vseng! Can you provide more information, like a link to the API documentation for this?

The approach of putting the system message directly on the model is not how any of the other LLM services work.

Thanks!

vseng:
Sure, @brainlid! Here is the documentation I referenced when looking into adding this functionality: [VertexAI Docs](https://cloud.google.com/vertex-ai/generative-ai/docs/learn/prompts/system-instructions#gemini-system-instructions-code-samples-drest)

They have example json here that shows where the system instruction should go (at the same level as "content" and "generation_config") It is admittedly different from how the other LLM services, including other Gemini services, work. 

Thanks so much for taking time to look this over!

brainlid:
Thank you for linking to the docs!

I think the smoothest way to support this is to the keep the
LangChain.Message with role :system, then, when building the API call, use
the Utils function that pulls off the system message from the list of
messages. It would submit the system prompt into the API documented field.

This way, you can take your saved conversation with ChatGPT and continue it
with Gemini and your application code doesn’t change. It also allows people
to still use PromptTemplates for system messages.

lib/chat_models/chat_vertex_ai.ex
```diff
@@ -128,8 +123,11 @@ defmodule LangChain.ChatModels.ChatVertexAI do
   end
 
   def for_api(%ChatVertexAI{} = vertex_ai, messages, functions) do
+    sys_instructions = messages |> Enum.find(fn x -> x.role == :system end) |> List.wrap()
+    other_messages = Enum.reject(messages, fn x -> x in sys_instructions end) |> List.wrap()
```
brainlid, line 127: This can be done using `LangChain.Utils.split_system_message/1`

```elixir
{sys_instructions, other_messages} = Utils.split_system_message(messages)
```

lib/chat_models/chat_vertex_ai.ex
```diff
@@ -144,9 +142,13 @@ defmodule LangChain.ChatModels.ChatVertexAI do
     }
 
     req =
-      if vertex_ai.system_instruction do
+      if length(sys_instructions) == 1 do
```
brainlid, line 145: This becomes...

```elixir
req = 
  if sys_instructions do
    Map.put(req, "system_instruction", for_api(sys_instructions))
  else
    req
  end
```

brainlid (changes requested):
I suggested some changes. This could be further streamlined by adding another `for_api/1` function.

```elixir
  defp for_api(nil), do: nil
```

Then `req` could be updated like this...

```elixir
    req = 
      %{
        "contents" => messages_for_api,
        "generationConfig" => %{
          "temperature" => vertex_ai.temperature,
          "topP" => vertex_ai.top_p,
          "topK" => vertex_ai.top_k
        }
      }
      |> Utils.conditionally_add_to_map("system_instruction", for_api(sys_instructions))
```

When you confirm it's working for you, please add some test coverage of the `ChatVertextAI` for this fixed behavior so we can prevent regressions.

Thanks for getting involved and helping to make this better for everyone!

vseng:
I've got it working on my end and have the test updated - thank you for the review!

brainlid:
Thanks @vseng! Looks great. Thanks for the updates and taking the time to make the library better. :grin: 
❤️💛💙💜

PR #262, opened 2025-03-07
feat: Custom errors fallback

yoschihirokaimoto:


lib/chains/llm_chain.ex
```diff
@@ -420,13 +420,31 @@ defmodule LangChain.Chains.LLMChain do
...
+
         {:error, _error_chain, reason} ->
           # run attempt received an error. Try again with the next LLM
           Logger.warning("LLM call failed, using next fallback. Reason: #{inspect(reason)}")
 
           try_chain_with_llm(use_chain, tail, before_fallback_fn, run_fn)
       end
     rescue
+      err in LangChainError ->
+        if err.type == :stop do
```
brainlid, line 436: The LangChainError `type` field is documented as being a string. In many cases, when we can recognize the type of error from the LLM/service, we populate it with a string representation here. 

However, you are comparing it to a `:stop` atom.

I like the idea of raising an intentional stop. I can see that being helpful. 

To support this:
- it should be a string
- a LangChainError function specifically for creating a stop would be helpful so it can be documented and exposed
- documentation around raising an stop exception, why you might want to, and what it does (ie: how the library handles it and responds)

brainlid (changes requested):
I like where I think you're going with raising a stop error, but would like for it to be more officially supported and documented.

Additionally, test coverage is needed for ensuring it isn't unintentionally broken in the future.

CaiqueMitsuoka:
Hey @brainlid thanks for the review. This open by mistake, it suppose to be opened to our fork haha.
I'm glad you liked, we are testing this in our app. The ideia is give control to tools so they can stop the chain and preserve the state. On some tools we have specific responses that we want to send instead of having the LLM process the rest. But it can be used to guardrails, abort, etc...
Lets make a good version of it with tests and docs and have make the contribution 💪

PR #266, opened 2025-03-14
Redact api-key from models

raulpe7eira:
### Summary

- **Problem:**

  If we inspect some models with API key, its possible to log and expose this sensitive data.

- **Fix:**

  Add a redact option on each sensitive field as described in [ecto doc](https://hexdocs.pm/ecto/Ecto.Schema.html#module-redacting-fields)


brainlid (approved):
Thanks for the contribution! A great addition.
❤️💛💙💜

PR #269, opened 2025-03-17
Support caching tool results for Anthropic calls

ci:
Anthropic supports the `"cache_control" => %{"type" => "ephemeral"}` block for anything under the content blocks in `messages.content` (based on [docs](https://docs.anthropic.com/en/docs/build-with-claude/prompt-caching#what-can-be-cached)).

Looks like currently only the text content in https://github.com/brainlid/langchain/blob/38e957d2985b054ce51e087b51bb1af54aee9754/lib/chat_models/chat_anthropic.ex#L872-L881 was supported

This adds the options keyword similar to how ContentPart implements it, to allow setting a tool result to be cached or not.

### DIFF
--- a/lib/chat_models/chat_anthropic.ex
+++ b/lib/chat_models/chat_anthropic.ex
@@ -102,6 +102,8 @@ defmodule LangChain.ChatModels.ChatAnthropic do
 
   @current_config_version 1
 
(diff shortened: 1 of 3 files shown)


brainlid:
Nice! Thanks @ci!
❤️💛💙💜

PR #273, opened 2025-03-18
Support choice of Anthropic beta headers

ci:
Adds support for specifying multiple betas, similar to their official SDKs.

(I was playing with `token-efficient-tools-2025-02-19` and long context, and used this to add it)


brainlid:
Thanks @ci! I appreciate your attention to the tests.
❤️💛💙💜

Is the `tools-2024-04-04` beta header even still in beta? I wonder if it's official now and could be removed.

ci:
Weirdly enough, they don't mention it anymore, but I tried without and got some failed live tests so just left it in, I thought about removing it too - maybe it was flaky though 🤔

PR #274, opened 2025-03-18
Fix OpenAI verbose_api

aaparmeggiani:
Quick fix to enable correct verbose_api for OpenAi


brainlid:
Thanks @aaparmeggiani!
❤️💛💙💜

PR #275, opened 2025-03-21
feat: add support for pdf content with OpenAI model

bwan-nan:
## What does this PR do?

OpenAI [now supports PDF files as input](https://platform.openai.com/docs/guides/pdf-files?api-mode=chat&utm_source=alphasigna), either as Base64-encoded data or as file IDs obtained after uploading files to the `/v1/files` endpoint.

This PR adds support for pdf content as Base64-encoded data.


brainlid:
Hi @bwan-nan! Thanks for the contribution and the inclusion of docs and tests!
❤️💛💙💜

klacointe:
Very useful, can you release a version with this feature? :heart_eyes:

PR #282, opened 2025-03-24
implement pdf for google ai

xrty-oss:


lib/chat_models/chat_google_ai.ex
```diff
@@ -311,6 +311,26 @@ defmodule LangChain.ChatModels.ChatGoogleAI do
...
+    mime_type =
+      case Keyword.get(opts || [], :media, nil) do
+        type when is_binary(type) ->
+          "application/#{type}"
+
+        other ->
+          message = "Received unsupported media type for ContentPart: #{inspect(other)}"
+          Logger.error(message)
+          raise LangChainError, message
+      end
```
brainlid, line 324: I'm a little confused why you are using `type when is_binary(type)` but the test is passing in the `:pdf` atom.

As you know (because you're using it), a ContentPart already supports a "media" type. So you could say `type: :file, media: "application/pdf"` and simplify this to some degree by not needing to match on supported or unsupported. It could just return the map data using whatever media mime type the developer provides.

Would that work for you? Or am I missing something?

brainlid (changes requested):
Thanks for working on this. I think it's close. I have a question and a suggestion.

lib/chat_models/chat_google_ai.ex
xrty-oss: you are missing nothing;  I started with one implementation and then found that the https://github.com/brainlid/langchain/pull/275 was now in, so I tried to quickly conform to that while keeping the work I am doing that requires this on track.

I will adjust as suggested. thank you.

brainlid:
Closed this in favor of #478

PR #283, opened 2025-03-25
Support for file with file_id in ChatOpenAI

nallwhy:
When creating a `ContentPart`, now determine how the file is handled based on `opts[:type]`.
For backward compatibility, the default value of `opts[:type]` is set to `:base64`.

### DIFF
--- a/lib/chat_models/chat_open_ai.ex
+++ b/lib/chat_models/chat_open_ai.ex
@@ -458,12 +458,23 @@ defmodule LangChain.ChatModels.ChatOpenAI do
   end
 
   def for_api(%_{} = _model, %ContentPart{type: :file, options: opts} = part) do
+    file_params =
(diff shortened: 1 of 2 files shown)


nallwhy:
Alternatively, we could consider adding a new `:uploaded_file` type to `ContentPart`, and handle it separately like `:image` or `:image_url`. That might also be a clean approach.

brainlid:
Hi @nallwhy! Sorry for my delay on this. I'd like to publish a new release soon before I publish a new major version with breaking changes.

This looks good. I think a documentation example would be helpful for people to know that this new feature exists without digging through code. Because it's an OpenAI specific handling, a section in the OpenAI module docs seems like a decent place for it.

brainlid:
Aw, why not. I'll merge it. But I'd still love to have some docs to cover it. :slightly_smiling_face: 
❤️💛💙💜

PR #287, opened 2025-03-31
support streaming responses from mistral

manukall:
This adds support for streaming responses from Mistral, plus live call tests.

There's sadly little documentation on how the API actually works, but it's supposedly "mostly compatible" with OpenAI's. In any case, it didn't work before, and works at least for the cases I tested with my changes.


brainlid:
Thanks for fixing that @manukall!
❤️💛💙💜

PR #295, opened 2025-04-22
Add thinking_mode, thinking_budget to ChatGoogleAI

nallwhy:


brainlid:
Hi @nallwhy! Oh, Google. :disappointed: 

What I do in my apps, is I can pattern match on a model name to get the Chat module setup I want. Like:

```elixir
def setup_model("anthropic" <> _rest = model) do
  # ...
end
```

Documenting an example of how people should use it would probably be the most helpful. Otherwise, as soon as there's a new gemini model, the library may be making the wrong choice and blocking them from what they need to do.

Add a basic test and some docs to the module doc so people know it's there and how to use it and we'll get it in!

brainlid:
@nallwhy Oh, BTW, I've got a branch that's about to become v0.4.0-rc.0 that introduces breaking changes in order to broadly add "thinking" support for models. 

My changes focus on Assistant's being able to provide ContentPart for their responses, and converting all received messages to a list of ContentPart to help make it more uniform.  

I only add support for OpenAI and Anthropic, but all the other functions and supporting code is updated.

Depending on how Gemini returns thinking blocks, you may want to use that approach.

nallwhy:
@brainlid Thanks! I’ll continue this work after 0.4.0 is released.
I’m planning to:
	•	Manage the application of thinking_mode based on model name, similar to how you use setup_model/1
	•	Convert thinking blocks into ContentPart for better consistency

nallwhy:
@brainlid 

I’m almost done, but I’m currently facing an issue when merging `MessageDelta` in the case of `stream: true`.
The following code should succeed, but there’s a problem when merging content parts of different types.

```elixir
deltas =
  [
    %LangChain.MessageDelta{
      content: %LangChain.Message.ContentPart{
        type: :thinking,
        content:
          "**Refining the Questions**\n\nI'm now zeroing in on the core philosophical questions. I'm expanding my initial list to include a deeper dive into the implications for ethics and agency in AI systems. The structure is taking form. I am exploring consciousness, ethics, agency, and personhood, as these are the main aspects.\n\n\n",
        options: []
      },
      merged_content: [],
      status: :incomplete,
      index: 0,
      role: :assistant,
      tool_calls: nil,
    },
    %LangChain.MessageDelta{
      content: %LangChain.Message.ContentPart{
        type: :text,
        content: "Okay, here is a deeply",
        options: []
      },
      merged_content: [],
      status: :incomplete,
      index: 0,
      role: :assistant,
      tool_calls: nil,
    }
  ]

LangChain.MessageDelta.merge_deltas(deltas)
```

```
11:58:33.281 [warning] Cannot merge content parts of different types: ...
```

Looking at `MessageDelta.merge_content_part_at_index/3`, it seems the position is determined via the index in the parts.
Does this mean that we don’t mix different content part types (e.g., thinking, text) in a single Message, but instead separate them into different `Messages` as a rule?

If that’s the case, when a `MessageDelta` with a different content type is streamed, I may need to manually mark the current message as `status: :complete` in `ChatGoogleAI`.

I’d like to hear your thoughts on the best way to handle this.
Let me know what you think.

nallwhy:
#354

PR #296, opened 2025-04-22
vertex ai file url support

ahsandar:
Vertex AI supports file url and can work with any file url compared to google ai which only  support file url uploaded using its files api, google ai support was added recently and most of the ground work is already there in https://github.com/brainlid/langchain/pull/286  to support file _url for content part in vertex .


brainlid:
Hi @ahsandar! It required a little cleanup, and I'm trusting you that it works for Vertex. Thanks for the test coverage and the contribution!
❤️💛💙💜

ahsandar:
Thanks. Will update further if required with follow up PRs

PR #301, opened 2025-04-30
Add strict to function of ChatModels.ChatOpenAI

nallwhy:
The `strict` field defined in `Function` was not being reflected in the resulting schema when passed to `ChatOpenAI`.


brainlid:
Hi @nallwhy! Sorry I've let this sit here so long. Thanks!
❤️💛💙💜

PR #309, opened 2025-05-24
Fix ContentPart migration

mathieuripert:
# Fix ContentPart Migration Compatibility Across Chat Models

### Problem
After the migration to ContentPart-based message content, 19 tests were failing because chat models weren't properly handling the new ContentPart list format. Models expected strings but received ContentPart structures.

### Changes
- Added missing ContentPart handling in ChatOllamaAI, ChatMistralAI, ChatPerplexity, ChatOpenAI, ChatVertexAI, and ChatGoogleAI
- Added extract_content_text/1 helpers to convert ContentPart lists back to strings where needed
- Fixed system message handling in ChatVertexAI and ChatGoogleAI to extract text from ContentPart structures
- Updated test expectations to use ContentPart.text!() format instead of plain strings
- Fixed role mapping in ChatVertexAI to use atoms consistently

All tests should pass now ✅

### DIFF
--- a/lib/chat_models/chat_google_ai.ex
+++ b/lib/chat_models/chat_google_ai.ex
@@ -194,7 +194,7 @@ defmodule LangChain.ChatModels.ChatGoogleAI do
           # Extract text from ContentPart structures
           text_content =
             content
-            |> Enum.filter(&match?(%LangChain.Message.ContentPart{type: :text}, &1))
+            |> Enum.filter(&match?(%ContentPart{type: :text}, &1))
             |> Enum.map(& &1.content)
             |> Enum.join(" ")
           %{"parts" => [%{"text" => text_content}]}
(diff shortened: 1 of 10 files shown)


mathieuripert:
@brainlid First of all, thanks for putting together this great library 👏 
Please check this PR whenever you can.

brainlid (changes requested):
Hi @mathieuripert! 

Thank you for taking up the challenge and helping to complete the rest of chat models! :grinning: 

Instead of introducing `extract_content_text`, can you try to use `ContentPart.parts_to_string/1`? I think it already covers the use cases you are handling here.

Thanks!

brainlid (approved):
I think it looks good! Thanks for the adjustments!

I feel a bit mixed by these: 
```elixir
    content = case msg.content do
      content when is_binary(content) -> content
      content when is_list(content) -> ContentPart.parts_to_string(content)
      nil -> nil
    end
```

Specifically that it expects the message.content to be a string instead of ContentParts. However, this works well for backward compatibility and if you aren't using a chat model with content parts then it makes sense.

Thanks!

brainlid:
Thanks @mathieuripert!
❤️💛💙💜

PR #310, opened 2025-05-25
feat: add configurable tokenizer to text splitters

mathieuripert:
## Problem
The text splitters (CharacterTextSplitter and RecursiveCharacterTextSplitter) currently use character count (String.length/1) for determining chunk sizes. This is good but in most cases we would like to use tokenizers that mimic what llms use (for example https://github.com/openai/tiktoken), so it would be good to pass custom tokenizer function instead.

## Proposed solution
- Added a configurable tokenizer field to both text splitter modules:
- Backward Compatible: Default tokenizer remains &String.length/1

Example usage:
```
# Custom word tokenizer
word_tokenizer = fn text -> text |> String.split() |> length() end
splitter = CharacterTextSplitter.new!(%{
  chunk_size: 50,  # 50 words instead of characters
  tokenizer: word_tokenizer
})
```

should resolve https://github.com/brainlid/langchain/issues/311

### DIFF
--- a/lib/text_splitter.ex
+++ b/lib/text_splitter.ex
@@ -11,8 +11,8 @@ defmodule LangChain.TextSplitter do
   end
 
   defp merge_split_helper(d, acc, text_splitter, separator) do
-    separator_len = String.length(separator)
-    len = String.length(d)
+    separator_len = text_splitter.tokenizer.(separator)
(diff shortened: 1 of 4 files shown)


brainlid:
Hi @mathieuripert! I think it looks good. Thanks!
❤️💛💙💜

Issue #313, opened 2025-06-01
Parsing OpenAI (compatible) endpoint response fails if `tools_calls` is `nil`


udoschneider:
Some self-hosted OpenAPI endpoints do return `nil` instead of an empty list `[]` for `"tool_calls"`. This breaks the code in
https://github.com/brainlid/langchain/blob/68e083411c8bfe2d722d610d0add215185384489/lib/chat_models/chat_open_ai.ex#L950-L954

Maybe using something like 
```elixir
"tool_calls" => Enum.map(calls || [], &do_process_response(model, &1))
```
at https://github.com/brainlid/langchain/blob/68e083411c8bfe2d722d610d0add215185384489/lib/chat_models/chat_open_ai.ex#L961 might be useful.

brainlid:
Thanks for your fix!

PR #314, opened 2025-06-01
Fix `tool_calls` `nil` messages

udoschneider:
Quick fix for #313.

Not sure if I understood the code structure correctly, but this fixes the issues I have.


brainlid:
Thanks @udoschneider! I encountered a similar issue with ChatAnthropic and fixed it there. Thanks for the PR!
❤️💛💙💜

PR #316, opened 2025-06-05
Fix: Add token usage to ChatGoogleAI message metadata

mathieuripert:
# Problem
ChatGoogleAI was inconsistent with other chat models in how it handled token usage information. While it extracted token usage from API responses and fired callbacks correctly, it wasn't adding the usage data to the message metadata like ChatOpenAI and ChatAnthropic do.

# Solution
- Modified do_process_response/3 in ChatGoogleAI to use TokenUsage.set/2 to add token usage to each processed message
- This follows the same pattern used by ChatOpenAI and ChatAnthropic


brainlid:
Thanks @mathieuripert! You rock!
❤️💛💙💜

PR #317, opened 2025-06-06
feat: include raw API responses in LLM error objects for better debug…

TwistingTwists:
- include the raw API responses in the original messages for debugging.


brainlid:
Thanks @TwistingTwists! This is a great enhancement. Thanks for your help!
❤️💛💙💜

PR #318, opened 2025-06-07
Support Open AI Responses API (w/ files, web search)

LannyBose:
Ok, this is _hard_ work-in-progress, but I'm to a point where I think I'm beyond getting things working and getting to actual decision-making, and I'd like some feedback or validation on steps forward.

Just to give a summary of where this PR is at so far:

### Working (?) so far:
- Text in. Text out.
- Images in. Text out.
- Base 64 encoded, non-image files in. Text out.
- In-text request for Native web search tool in. Annotated web search text out.

### Definitely remaining to be addressed:
- The Responses API can return "refusals", which are fundamentally just text. I'm currently throwing them out. We should emit them as text content both with and without `stream: true`. Straightforward, just not done yet.
- Reasoning. I haven't yet looked at how this is handled in other `ChatModels`
- Token-usage gathering (I copy-pasted from `ChatModels.ChatOpenAI`, just haven't checked to see if it's working as-is)
- Checking to make sure usage headers didn't change (same)
- Deciding which `NativeTool`s other than `web_search_preview` get to make the cut (I have opinions, but it's a worthwhile discussion)
  - `file_search_call`
  - `computer_use_preview`
  - `mcp`
  - `code_interpreter`
  - `image_generation`
  - `local_shell`
- Spec-writing

### Decisions to be validated:
- Big one. Does making this its own `ChatModel.OpenAIResponses` make sense? It did to me since the input structure and output structure of the Responses API are distant-enough from Chat Completions to necessitate its own thing.
- [Open AI has a files endpoint](https://platform.openai.com/docs/guides/pdf-files?api-mode=responses), where you can separately upload a file and send its ID as a content part. I allow for the IDs to be passed in as `content` but I'm not touching the actual upload in that case.
- [The Responses API has a facility to send a single new message and `previous_response_id`](https://platform.openai.com/docs/api-reference/responses/create#responses-create-previous_response_id), removing the need to send the whole chain. I assume our holding-of-the-chain would mean we're not interested in that.
- [The responses API can emit a bunch of `content_items`](https://platform.openai.com/docs/api-reference/responses/object#responses/object-output) (a message, a tool_call, another message) each of which may contain one or more `content_parts` (text, refusal).
  - Because an overall response _could_ return an arbitrary number of content_items, I _could_ see us treating them as separate `LLMChain.Message`s, but  (I believe) having a `do_process_response` return multiple messages is meant to indicate `choices` and not multiple sequential messages being returned.
- I'm "throwing out" [a lot of the emitted events](https://platform.openai.com/docs/api-reference/responses-streaming) when `stream: true`. It would be nice to get extra eyes on it to ensure I'm not missing events that provide parallel functionality to other LLMs in this package.

### DIFF
--- a/lib/chat_models/chat_open_ai_responses.ex
+++ b/lib/chat_models/chat_open_ai_responses.ex
@@ -1,0 +1,1207 @@
+defmodule LangChain.ChatModels.ChatOpenAIResponses do
+  @moduledoc """
+  Represents the OpenAI Responses API
+
+  Parses and validates inputs for making requests to the OpenAI Responses API.
+
+  Converts responses into more specialized `LangChain` data structures.
+
+  ## ContentPart Types
+
+  OpenAI's Responses API supports several types of content parts that can be combined in a single message:
+
+  ### Text Content
+  Basic text content is the default and most common type:
+
+      Message.new_user!("Hello, how are you?")
+
+  ### Image Content
+  OpenAI supports both base64-encoded images and image URLs:
+
+      # Using a base64 encoded image
+      Message.new_user!([
+        ContentPart.text!("What's in this image?"),
+        ContentPart.image!("base64_encoded_image_data", media: :jpg)
+      ])
+
+      # Using an image URL
+      Message.new_user!([
+        ContentPart.text!("Describe this image:"),
+        ContentPart.image_url!("https://example.com/image.jpg")
+      ])
+
+  For images, you can specify the detail level which affects token usage:
+  - `detail: "low"` - Lower resolution, fewer tokens
+  - `detail: "high"` - Higher resolution, more tokens
+  - `detail: "auto"` - Let the model decide
+
+  ### File Content
+  OpenAI supports both base64-encoded files and file IDs:
+
+      # Using a base64 encoded file
+      Message.new_user!([
+        ContentPart.text!("Process this file:"),
+        ContentPart.file!("base64_encoded_file_data",
+          type: :base64,
+          filename: "document.pdf"
+        )
+      ])
+
+      # Using a file ID (after uploading to OpenAI)
+      Message.new_user!([
+        ContentPart.text!("Process this file:"),
+        ContentPart.file!("file-1234", type: :file_id)
+      ])
+
+  ## Callbacks
+
+  See the set of available callbacks: `LangChain.Chains.ChainCallbacks`
+
+  ### Rate Limit API Response Headers
+
+  OpenAI returns rate limit information in the response headers. Those can be
+  accessed using the LLM callback `on_llm_ratelimit_info` like this:
+
+      handlers = %{
+        on_llm_ratelimit_info: fn _model, headers ->
+          IO.inspect(headers)
+        end
+      }
+
+      {:ok, chat} = ChatOpenAI.new(%{callbacks: [handlers]})
+
+  When a request is received, something similar to the following will be output
+  to the console.
+
+      %{
+        "x-ratelimit-limit-requests" => ["5000"],
+        "x-ratelimit-limit-tokens" => ["160000"],
+        "x-ratelimit-remaining-requests" => ["4999"],
+        "x-ratelimit-remaining-tokens" => ["159973"],
+        "x-ratelimit-reset-requests" => ["12ms"],
+        "x-ratelimit-reset-tokens" => ["10ms"],
+        "x-request-id" => ["req_1234"]
+      }
+
+  ### Token Usage
+
+  OpenAI returns token usage information as part of the response body. The
+  `LangChain.TokenUsage` is added to the `metadata` of the `LangChain.Message`
+  and `LangChain.MessageDelta` structs that are processed under the `:usage`
+  key.
+
+  The OpenAI documentation instructs to provide the `stream_options` with the
+  `include_usage: true` for the information to be provided.
+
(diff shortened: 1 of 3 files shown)


lib/chat_models/chat_open_ai_responses.ex
```diff
@@ -0,0 +1,1196 @@
...
+        search_context_size: "medium",
+        user_location: %{
+          type: "approximate",
+          city: "Humble",
+          country: "US",
+          region: "Texas",
+          timezone: "America/Chicago"
+        }
+      }
+      native_web_tool = NativeTool.new!(%{name: "web_search_preview", configuration: web_config)
```
LannyBose, line 131: This choice of using `NativeTool` in this way needs some validation

lib/chat_models/chat_open_ai_responses.ex
```diff
@@ -0,0 +1,1196 @@
...
+    field :endpoint, :string, default: "https://api.openai.com/v1/responses"
+
+    field :model, :string, default: "gpt-3.5-turbo"
+
+    field :include, {:array, :string}, default: []
+    # omit instructions becasue langchain assumes statelessness
+    field :max_output_tokens, :integer, default: nil
+    # omit metadata because chat_open_ai also omits it
+    # omit parallel_tool_calls because chat_open_ai also omits it
+    # omit previous_response_id becasue langchain assumes statelessness
```
LannyBose, line 218: Choice needing validation

lib/chat_models/chat_open_ai_responses.ex
```diff
@@ -0,0 +1,1196 @@
...
+  end
+
+  @doc """
+  Convert a ContentPart to the expected map of data for the OpenAI API.
+  """
+  def content_part_for_api(%ChatOpenAIResponses{} = _model, %ContentPart{type: :text} = part) do
+    %{"type" => "input_text", "text" => part.content}
+  end
+
+  def content_part_for_api(
```
LannyBose, line 534: I wonder if this treatment of file options feels idiomatic compared to other `ChatModels`

brainlid:
I started looking more into the responses API and found this document:
- https://platform.openai.com/docs/guides/responses-vs-chat-completions

It's a helpful starting point for me.

I'm looking into add built-in tool support for web_search which Anthropic and OpenAI both support, so I'm looking for commonalities and a unified representation.

But yes, it makes sense to be a separate ChatModel as the differences from responses to chat are fundamental.

brainlid:
As for the [previous_response_id](https://platform.openai.com/docs/api-reference/responses/create#responses-create-previous_response_id), Anthropic has something similar which they call caching. I'd like to add a message-level attribute for signaling that caching/saving should happen at that point. 

I think this means the message.id needs to be kept from the what the LLM API returns. I don't think this is happening now.

Then, perhaps internally within the chain, we _optionally_ still manage the entire message chain, but LLMChain looks for that cached/saved ID and only sends from that point on? 

With Anthropic, if the cache (which truly is a cache and is remove after some time) is invalid, then the LLMChain can handle that response and resend the whole chain. If it doesn't have the whole chain (user opted not to include the messages prior to the save point), then it errors as it can't be resumed.

brainlid:
Do you mind if I hack on this PR branch too?

LannyBose:
@brainlid By all means! I think the approach towards previous_response_id works, though I wonder how to signal what provider the id belongs to? In the case of a fallback LLM, we need to know that the previous_response_id isn't relevant and that we need to send the whole chain.

brainlid:
@LannyBose, If you fallback to ChatGPT on Azure, then it's true, it won't have that data stored. We'll have to see what kind of error messages are received when referring to an invalid reference. If we can detect that, then possibly the chain can auto-retry with the whole conversation?

brainlid:
@LannyBose I'm currently looking into NativeTool support with Anthropic while keeping an eye on how OpenAI responses handles it. I'll let you know when I have something ready. But I'm hoping to find a way to use ContentParts for storing the annotations.

lib/chat_models/chat_open_ai_responses.ex
```diff
@@ -0,0 +1,1197 @@
...
+    |> Req.post(
+      into: Utils.handle_stream_fn(openai, &decode_stream/1, &do_process_response(openai, &1))
+    )
+    |> case do
+      {:ok, %Req.Response{body: data} = response} ->
+        Callbacks.fire(openai.callbacks, :on_llm_ratelimit_info, [
+          get_ratelimit_info(response.headers)
+        ])
+
+        data
```
arjan, line 777: Responses list should be flattened I think,

```elixir
List.flatten(data)
```

brainlid:
Hi @LannyBose! This is still marked as a draft. What would you like from me to help move this forward?

arjan:
I have continued on this PR on my own branch here: https://github.com/arjan/langchain/tree/openai-responses

Not sure if 2 separate modules is the way to go, or that we should go the route of the other PR: https://github.com/brainlid/langchain/pull/369

vasspilka:
I closed my PR since I feel this one is in better shape and @brainlid is more familiar with it

LannyBose:
Hey all! Sorry, other app-code took focus. Going to circle back and see where I can contribute next

arjan:
This ticket can be closed, as the fork-of-the-fork of this PR has been merged.

LannyBose:
@arjan Yep! TY for the nudge

PR #319, opened 2025-06-10
Vertex AI Native tool calling

ahsandar:
Adding support for NativeTool to Vertex AI to make it at par with Google AI implementation in Langchain.


FeeJai:
may I ask the reason this was closed?

PR #321, opened 2025-06-13
Add structured output support for Anthropic with json_response

mathieuripert:
## Add structured output support for Anthropic with json_response
This PR adds support for structured JSON output with Anthropic's Claude models by implementing the recommended approach using tool calls. While Anthropic doesn't support a direct "JSON mode" like some other LLMs, they officially recommend using tool calls for structured output.

https://docs.anthropic.com/en/docs/agents-and-tools/tool-use/overview#json-mode

### Details
Added support for json_response: true and json_schema parameters in ChatAnthropic
When these parameters are set, the implementation:
Creates a special structured_output tool with the provided schema
Forces tool choice to use this tool
Extracts the tool call response and converts it to a properly formatted JSON string
Returns the JSON string as a ContentPart in the message's content field

### DIFF
--- a/lib/chat_models/chat_anthropic.ex
+++ b/lib/chat_models/chat_anthropic.ex
@@ -230,6 +230,10 @@ defmodule LangChain.ChatModels.ChatAnthropic do
     # Tool choice option
     field :tool_choice, :map
 
+    # JSON response format fields for structured output
+    field :json_response, :boolean, default: false
+    field :json_schema, :map, default: nil
+
     # Beta headers
     # https://docs.anthropic.com/claude/docs/tool-use - requires tools-2024-04-04 header during beta
     field :beta_headers, {:array, :string}, default: ["tools-2024-04-04"]
@@ -253,6 +257,8 @@ defmodule LangChain.ChatModels.ChatAnthropic do
     :stream,
     :thinking,
     :tool_choice,
+    :json_response,
+    :json_schema,
     :beta_headers,
     :verbose_api
   ]
@@ -322,6 +328,26 @@ defmodule LangChain.ChatModels.ChatAnthropic do
       |> Enum.map(&message_for_api/1)
       |> post_process_and_combine_messages()
 
+    # Handle structured output by creating a tool when json_response is enabled
+    {api_tools, api_tool_choice} =
+      case {get_structured_output_tool(anthropic), tools} do
+        {nil, tools} ->
+          # No structured output, use regular tools and tool_choice
+          {get_tools_for_api(tools), get_tool_choice(anthropic)}
+
+        {structured_tool, []} ->
+          # Structured output only, no regular tools
+          {[structured_tool], %{"type" => "tool", "name" => "structured_output"}}
+
+        {structured_tool, _tools} ->
+          # Both structured output and regular tools - prioritize structured output
+          Logger.warning(
+            "Both json_response and tools provided. Structured output will take precedence."
+          )
+
+          {[structured_tool], %{"type" => "tool", "name" => "structured_output"}}
+      end
+
     %{
       model: anthropic.model,
       temperature: anthropic.temperature,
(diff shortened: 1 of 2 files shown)


brainlid:
Hi @mathieuripert!

Thanks for the PR! I'd like to understand a bit more of what you see as the benefit or need for this change?

Currently, specifying the `tool_choice` effectively does this, while also making it easier for the developer to provide a specific tool with a custom name and description (which factors into the LLMs behavior).

Is the goal to make the shape of a ChatOpenAI and ChatAnthropic model be the same with `json_mode`? Or is there another use-case you have in mind?

mathieuripert:
@brainlid 
Yes my initial goal was to create an unified interface between the different models to be able to swap models easily inside my system. But I get your point and it might not be realistic to make all models the same shape given they all have their own specifics and custom parameters. So yes it may not worth the effort and probably better to keep it more explicit and the responsibility of the developper to call each model differently.

mathieuripert:
So I am good to close this PR !

brainlid:
Thanks @mathieuripert for confirming your goal! I was thinking more about it and I can see the merit of what you're thinking. 

Here's how I've been thinking about:
- the ChatModel's define how to represent and work with a given LLM/service.
- the LLMChain expresses the desire of what to do (send messages, receive tool calls, etc) and the ChatModel is really the adapter.

I'd like to keep the API attributes of the ChatModel matching what the service defines.

But, the idea of an enforced JSON response and a specific tool call usage is a commonly needed pattern. I started to think about expressing that use-case as a ChatModel behaviour that the LLMChain could call. This means the LLMChain is what's expressing the desired behavior for a data extraction-type call.

But, reviewing the behaviour `LangChain.ChatModels.ChatModel`, I'm not sure if it makes sense.

The primary behaviour function for the LLMChain calls:

```elixir
  @callback call(
              t(),
              String.t() | [Message.t()],
              [LangChain.Function.t()]
            ) :: call_response()
```

Does this mean I'm proposing a new callback like this?

```elixir
  @callback json_enforced_call(t(), schema(), String.t() | [Message.t()],
              [LangChain.Function.t()]
            ) :: call_response()
```

But that doesn't address when a specific function must be called. Would there be an option or a separate callback to enforce that a specific function is called? That starts to feel really messy. 

Still thinking. :thinking:

brainlid:
Thinking more about it, it's the combo a json enforced and a specific function that is forced to be called. This combo seems pretty exclusively to be a "data extraction" use-case. Do you feel that's what this is all really about?

If so, then I can possibly see a model behaviour around that use case. But, I think it may go better on the `LLMChain` because it could also verify that there are no other tools or processors included. Adding additional tools and message processors to a forced tool-use data extraction use-case seems to be a common mistake as well.

What do you think?

mathieuripert:
Hi @brainlid 
Thanks for thinking about it. 

Yes, what i am trying to achieve here with Anthropic is to force the use of a tool and the arguments of the tool are the response schema i want to output. So yes combo `json enforced and a specific function that is forced to be called`

And I agree with you, this is more a chain responsibility, as we need to extract the call arguments from llm response. 
However, the enforcement of the tool is passed to ChatModel with `tool_choice: %{"type" => "tool", "name" => tool_name}`

mathieuripert:
what i was thinking we could do is something in the spirit of 



```elixir

  def run_with_structured_output(chain, response_schema) do
    # construct a tool from a response_schema
    tool = structured_output_tool(response_schema)

    # Update the LLM in the chain to include tool_choice
    updated_llm = %{chain.llm | tool_choice: %{"type" => "tool", "name" => tool.name}}
    updated_chain = %{chain | llm: updated_llm}

    updated_chain
    |> LLMChain.add_tools([tool])
    |> LLMChain.run()
    |> do_process_tool_call_response()
  end
  
  defp do_process_tool_call_response({:ok, %LLMChain{last_message: last_message}}) do
    case last_message.tool_calls do
      [%{arguments: arguments}] ->
        arguments =
          arguments |> Enum.map(fn {k, v} -> {String.to_atom(k), v} end) |> Enum.into(%{})

        {:ok, arguments}

      [] ->
        {:error, "No tool calls found in response"}

      multiple when is_list(multiple) ->
        {:error, "Multiple tool calls found, expected exactly one"}
    end
  end
```

PR #324, opened 2025-06-16
feat(chat_perplexity): capture full API response metadata

mathieuripert:
### Capture Full Perplexity API Response Metadata

It enhances `ChatPerplexity` to capture the complete API response from Perplexity, not just the message content.
The main benefit is to get the citations and search results.

- Full API response metadata is now available in `Message.processed_content`
- Access to citations, search results, token usage, and response identifiers

### DIFF
--- a/lib/chat_models/chat_perplexity.ex
+++ b/lib/chat_models/chat_perplexity.ex
@@ -721,15 +721,19 @@ defmodule LangChain.ChatModels.ChatPerplexity do
     end
   end
 
-  def do_process_response(_model, %{
-        "choices" => [
-          %{
-            "delta" => %{"role" => role, "content" => content},
-            "finish_reason" => finish,
-            "index" => index
-          } = _choice
-        ]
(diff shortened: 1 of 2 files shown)


brainlid:
Hi @mathieuripert! Using processed_content as a place to store the additional information is an interesting approach!

One of the things I'm working on currently is support for NativeTools. This means web_search tools, which are also returning citations or annotations. The challenge is the variety of ways services return this data. But I'm looking for something more fully supported in ContentParts. I don't have it done yet, but I'll think more about it. 

Thanks!

mathieuripert:
Hi @brainlid, 
In the meantime, do you think we can merge this change ? This would allow the perplexity sources to be available in the response message. I think this is a simple workaround.
Thanks

brainlid:
Sorry this was left so long. But I just merged #461 which adds official citation support across multiple models, Perplexity being one of them. 

It'd love to hear how it works for you!

PR #327, opened 2025-06-24
Fix AWS Bedrock stream decoder ordering issue

stevehodgkiss:
When there are multiple messages in the buffer (`remaining != ""`), they're added to the list of chunks with `[chunk | chunks]`. That list wasn't being reversed before being returned, which results in incorrect output like `[" there are a", "Yes", "few concerning"]`.


anthonynovatsis (approved):

brainlid:
Thanks @stevehodgkiss! Great catch and fix! I appreciate the test coverage as well.
❤️💛💙💜

PR #329, opened 2025-07-07
Add `retry: transient` to Req for Anthropic models in stream mode

jonator:
If in streaming mode for Anthropic models, old connections would often close, causing an error. This leans on Req's built in retry mechanism to perform the default number of attempts to reopen the connection before throwing an error. This mechanism is in place for non streaming mode, but not streaming mode in the Anthropic Chat Model.


brainlid:
Hi @jonator! Excellent catch! Thanks!
❤️💛💙💜

PR #332, opened 2025-07-08
fix: Require gettext ~> 0.26

mweidner037:
The change to gettext's setup in https://github.com/brainlid/langchain/pull/271 requires 0.26 or higher, so we should reflect this in mix.exs.

Note that this will expose projects using langchain to the same warning fixed in that PR.

Context: https://hexdocs.pm/gettext/changelog.html#v0-26-0


brainlid:
Good call. Thanks @mweidner037! And thanks for doing the reference work to make this an easy accept. :slightly_smiling_face: 
❤️💛💙💜

PR #335, opened 2025-07-09
Added usage data to the VertexAI Message response.

raulchedrese:
Currently ChatVertexAI doesn't support returning token usage data. This change adds a TokenUsage struct to the Message response.

I tried to closely follow the implementation in ChatGoogleAI, which already has this feature.

### DIFF
--- a/.github/workflows/elixir.yml
+++ b/.github/workflows/elixir.yml
@@ -23,6 +23,7 @@ env:
   PERPLEXITY_API_KEY: invalid
   MISTRAL_API_KEY: invalid
   VERTEX_API_KEY: invalid
+  VERTEX_API_ENDPOINT: invalid
 
 permissions:
   contents: read


lib/chat_models/chat_vertex_ai.ex
```diff
@@ -482,9 +482,13 @@ defmodule LangChain.ChatModels.ChatVertexAI do
 
   def do_process_response(response, message_type \\ Message)
 
-  def do_process_response(%{"candidates" => candidates}, message_type) when is_list(candidates) do
+  def do_process_response(%{"candidates" => candidates} = data, message_type)
+      when is_list(candidates) do
+    token_usage = get_token_usage(data)
+
```
brainlid, line 488: After determining that we have token usage, it should fire a callback with the information:

```elixir
Callbacks.fire(model.callbacks, :on_llm_token_usage, [token_usage])
```

You can find numerous examples in the code when you search for `:on_llm_token_usage`.

I'd love to see test coverage of firing the event.

brainlid (changes requested):
From looking at the code, it would seem that it doesn't get token usage information from deltas, is that correct?

raulchedrese:
@brainlid I've added a couple of live tests to verify that the callback is being fired.

test/chat_models/chat_vertex_ai_test.exs
```diff
@@ -456,4 +505,102 @@ defmodule ChatModels.ChatVertexAITest do
...
+          send(self(), {:fired_token_usage, usage})
+        end
+      }
+
+      chat =
+        ChatVertexAI.new!(%{
+          model: "gemini-2.5-flash",
+          temperature: 0,
+          endpoint:
+            "https://aiplatform.googleapis.com/v1/projects/eng-sandbox-7s/locations/global/publishers/google",
```
brainlid, line 523: Is this endpoint something you want to include? It looks like it might be specific to you? If so, this could pull in an ENV for the endpoint. If there's a standard API endpoint available for use, it could be that one.

brainlid (commented):
Looks good! 

I just have a question about the live test setup and the endpoint URL.

Thanks!

brainlid:
Great! Thanks @raulchedrese!
❤️💛💙💜

PR #336, opened 2025-07-09
Add OpenAI Deep Research integration

fbettag:
Implement comprehensive deep research functionality using OpenAI's o3-deep-research and o4-mini-deep-research models. This integration provides powerful research capabilities that can analyze hundreds of sources to create detailed reports with citations.

Features:
- Full OpenAI Deep Research API integration
- Support for both o3-deep-research and o4-mini-deep-research models
- Async polling for long-running research (5-30 minutes)
- Automatic citation extraction and formatting
- Seamless LLMChain integration with existing tool patterns
- Comprehensive test coverage with live API verification
- Complete documentation and usage examples

The implementation follows existing LangChain patterns and reuses the same OpenAI API key configuration for consistency.

Also this has mostly been coded by Claude Code, i just made the plan really good and then did some testing myself (using the lib from my repo).

### DIFF
--- a/lib/tools/deep_research_client.ex
+++ b/lib/tools/deep_research_client.ex
@@ -157,7 +157,10 @@ defmodule LangChain.Tools.DeepResearchClient do
     # Build tools array based on include_code_interpreter
     tools =
       if include_code_interpreter do
-        [%{type: "web_search_preview"}, %{type: "code_interpreter"}]
+        [
+          %{type: "web_search_preview"},
+          %{type: "code_interpreter", container: %{type: "auto", file_ids: []}}
+        ]
       else
         [%{type: "web_search_preview"}]
       end


fbettag:
FYI i'm doing some more testing on a rather large project where i will replace Perplexity with this. But generally having a good feeling about this. Will take a couple of days, but feel free to test drive.

fbettag:
I tested this out all morning and blew 80$ in OpenAI tokens :] works pretty nicely. You just need to up the async_tool_timeout and i've just updated the documentation for it.

brainlid:
> I tested this out all morning and blew 80$ in OpenAI tokens :] works pretty nicely. You just need to up the async_tool_timeout and i've just updated the documentation for it.

:open_mouth: 

Wow. Yeah, working with and testing the deep research is a lot pricier!

With an expected duration ranging from 5 minutes to 30 minutes, I think it makes sense to show an example of how to get the research request ID back and possibly save it to a DB. I would hate for a long-running request to be lost because the app was deployed during that window.

Also, some of the tests can use mocking to test more of the library code while avoiding the cost and delay of a live test.

And finally, I want to make sure we can have these live tests excluded from a normal live_call test run. Sometimes when making a new release, I'll run `mix test --include live_call`, but I wouldn't want deep research to be part of that. So perhaps some notes and a different tag.

But the code, tests and documentation look good! I really appreciate your work and this contribution! :medal_sports:

brainlid:
I can see that your work supports "background mode". https://platform.openai.com/docs/guides/background

I didn't see a documentation example of this use-case, which I expect may be the default way people will use it.

I'm happy to help with a mocking setup if that's helpful for you too.

fbettag:
I don't think people themselves will use it much :D Since it's a tool-call it's up to the AI Agent to decide on model+background mode ;)

lib/tools/deep_research_client.ex
```diff
@@ -0,0 +1,324 @@
...
+  2. Checking request status
+  3. Retrieving completed results
+
+  Uses the same authentication and HTTP client patterns as other OpenAI integrations
+  in LangChain, reusing the existing `:openai_key` configuration.
+  """
+  require Logger
+  alias LangChain.Config
+
+  @base_url "https://api.openai.com/v1/responses"
```
jonator, line 16: IMO it may be useful to provide an opt to allow this to be overridden.

This would then support the scenario where someone is using deep research via an OpenAI API compatible provider like OpenRouter.

brainlid: Or via Azure (I assume it's available there, but not certain)

fbettag:
While using the lib extensively for the past week or so, i also noticed that it might make more sense to make background always true, just because of tool call timeouts that need to be adjusted if background is false and it's waiting for the result.

lib/tools/deep_research_client.ex
fbettag: added now

fbettag:
ping :)

brainlid:
Awesome @fbettag! Thanks for the great work. Great test coverage, updated docs, and the notebook is a really nice touch. :slightly_smiling_face:
❤️💛💙💜

And thanks for the gentle ping. :grin:

PR #337, opened 2025-07-09
feat: Add organization ID as a parameter

hjemmel:
Add support for organization ID as a parameter in the OpenAI chat model. This allows users to specify the organization context when making requests, enabling better management of resources and access control within the OpenAI API.
The change includes updates to the `ChatOpenAI` module and its tests to ensure the organization ID is correctly passed and utilized in API calls.

### DIFF
--- a/lib/chat_models/chat_open_ai.ex
+++ b/lib/chat_models/chat_open_ai.ex
@@ -206,6 +206,10 @@ defmodule LangChain.ChatModels.ChatOpenAI do
     # customer to provide their own.
     field :api_key, :string, redact: true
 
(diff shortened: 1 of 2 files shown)


brainlid:
Awesome @hjemmel! A solid addition.
❤️💛💙💜

brainlid:
I just realized that this was committed against v0.3.x. That's fine, but that's not the active branch. I merged the change into main.

hjemmel:
> I just realized that this was committed against v0.3.x. That's fine, but that's not the active branch. I merged the change into main.

Ops sorry @brainlid, I thought that branch was for the current release, my bad :(

PR #340, opened 2025-07-15
feat: Add image and file template support to PromptTemplate

marot:
Hi!

I was working on a project using AshAi and tried to send images to the AI provider. There is a snippet in AshAi:

```elixir
run prompt(
  ChatOpenAI.new!(%{model: "gpt-4o"}),
  prompt: [
    Message.new_system!("You are an expert at image analysis"),
    Message.new_user!([
      PromptTemplate.from_template!("Extra context: <%= @input.arguments.context %>"),
      ContentPart.image!("<%= @input.arguments.image_data %>", media: :jpg, detail: "low")
    ])
  ]
)
```

Since the intention of Ash is to be declarative, this is the desired way of declaring a prompt inside of an action. However, it doesn't actually work. The template in ContentPart.image! is taken as a string.

This PR adds support for `PromptTemplate.image_template!` and PromptTemplate.file_template!` so that it is possible to specify content parts with different types.

Please let me know if I should bring this up in an issue first.

Best,
MArco

### DIFF
--- a/lib/prompt_template.ex
+++ b/lib/prompt_template.ex
@@ -40,11 +40,13 @@ defmodule LangChain.PromptTemplate do
     field :text, :string
     field :inputs, :map, virtual: true, default: %{}
     field :role, Ecto.Enum, values: [:system, :user, :assistant, :function], default: :user
+    field :content_type, Ecto.Enum, values: [:text, :image, :file], default: :text
+    field :options, :any, virtual: true, default: []
   end
 
   @type t :: %PromptTemplate{}
 
-  @create_fields [:role, :text, :inputs]
+  @create_fields [:role, :text, :inputs, :content_type, :options]
   @required_fields [:text]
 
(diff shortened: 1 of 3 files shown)


lib/prompt_template.ex
```diff
@@ -339,9 +393,15 @@ defmodule LangChain.PromptTemplate do
           [Message.t()] | no_return()
   def to_messages!(prompts, %{} = inputs \\ %{}) when is_list(prompts) do
     Enum.map(prompts, fn
-      %PromptTemplate{} = template ->
+      %PromptTemplate{content_type: content_type} = template when content_type in [nil, :text] ->
         to_message!(template, inputs)
 
+      %PromptTemplate{content_type: content_type} = template
+      when content_type in [:image, :file] ->
+        # For image/file templates, create a user message with the content part
```
marot, line 401: Here I'm not sure if it makes sense to have images/files that don't have the user role

PR #342, opened 2025-07-21
Fix ChatOllamaAI stop sequences: change from string to array type

gur-xyz:
## Summary

  Fixes the `stop` field in `ChatOllamaAI` to match Ollama API requirements by changing type from `:string` to `{:array, :string}`. This resolves a critical bug where stop sequences were completely non-functional due to API type mismatch.

  ## Problem

  The original implementation used `field :stop, :string`, but Ollama API requires stop sequences as arrays and rejects strings with the error: `{"error":"option \"stop\" must be of type array"}`. This made the stop sequences feature completely
  unusable.

  ## Solution

  - **Changed field type**: `:string` → `{:array, :string}, default: []`
  - **Added smart filtering**: Empty arrays are excluded from API requests via `Utils.conditionally_add_to_map` to preserve modelfile defaults (prevents overriding built-in stop tokens like `<|eot_id|>`)
  - **Updated API conversion**: `for_api/3` now conditionally includes stop sequences only when non-empty

  ## Breaking Change

  This is a breaking change for users who were passing string values (though they would have received API errors). Users must now pass arrays:

  ```elixir
  # Before (broken):
  ChatOllamaAI.new!(%{"stop" => "\\n"})  # ❌ API error

  # After (working):
  ChatOllamaAI.new!(%{"stop" => ["\\n", "Human:"]})  # ✅ Works
```
  API Reference:

  [ollama api reference](https://github.com/ollama/ollama/blob/main/docs/api.md#generate-request-with-options)

  Tests

  Comprehensive test coverage added:

  - Basic functionality: Array-based stop sequences work correctly
  - Empty array handling: Empty arrays excluded from API (preserves modelfile defaults)
  - Input validation: Strings rejected, arrays accepted, nil handled gracefully
  - Edge cases: Unicode characters, special chars, escape sequences
  - Complex scenarios: Overlapping sequences, large arrays (100+ items)
  - API compatibility: Verifies correct JSON format sent to Ollama

  Documentation

  - Added inline code documentation explaining empty array filtering behavior
  - Updated CHANGELOG.md with breaking change details
  - Comprehensive test descriptions for maintainability

  Resolves the TODO comment in chat_ollama_ai_test.exs:103 and makes stop sequences fully functional for Ollama integration.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## [Unreleased]
+
+### Breaking Changes
+- **ChatOllamaAI**: Fixed `stop` field type from `:string` to `{:array, :string}` to match Ollama API requirements. Previously, stop sequences were non-functional due to API type mismatch. Now accepts arrays like `["\\n", "Human:", "<|eot_id|>"]`. Empty arrays are excluded from API requests to preserve modelfile defaults.
+
+---
+
 ## v0.4.0-rc.1
 
 ---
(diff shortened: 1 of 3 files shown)


brainlid:
Thanks for the background and the fix!
❤️💛💙💜

Issue #293, opened 2025-04-17
Proposal: Separate the Step Between Tool Result and Assistant Message


nallwhy:
I’d like to propose separating the step between the tool result and the assistant message in this library.

Currently, even when running in the default mode, after `execute_tool_calls/1` is called, the next step `do_run/1` is executed immediately, making it impossible for me to intervene or customize anything in between.

My use case is that when the `tool_result` returns a result for routing, I want to directly use this data (specifically the `processed_content`) to render a custom UI for the user at that point. This approach would allow me to provide a much faster and richer user experience by returning a tailored UI immediately after receiving the `tool_result`.

In my view, the entire flow—`message` -> `tool_call` -> `execute (tool result)` -> `message`—should be treated as a sequence of equal steps. If we make these steps equally accessible, users of this library will gain much more control over each stage, enabling a variety of advanced use cases.

Looking forward, I also hope to add support for the `while_needs_response` mode, where a function could be provided to determine at each step whether to proceed or pause, giving library users even more flexibility.

I believe that to achieve this, the library needs to treat each step in the chain as an equal unit, as described above. If the proposal is accepted, I am willing to work on this myself and contribute the changes to the library.

Thank you for considering this enhancement!

nallwhy:
I think https://github.com/brainlid/langchain/commit/459a392cb3bceb73b6d2f6980c885b52c9ec301c might address the issue I was facing. I’ll test it out and report back.

nallwhy:
After checking the changes, I realized that the new behavior always retries until the specific tool is used. It doesn’t work for my use case, as I also need to support generic tool usage.

brainlid:
Hi @nallwhy!

I have a couple thoughts. 

If you use the `run` without a mode, it stops at each received message. At this point you could make any adjustments or determinations then optionally continue the chain.

As for your rich UI idea, I do this by including a process to notify in the context that a function/tool receives. It could be a LiveView, PubSub, whatever. But it allows me to async reflect what's going on.

nallwhy:
Thanks for the reply!

As you suggested, I’m currently using run without a mode and handling the control flow myself. It did take a bit of work to implement the logic for when to continue and when to intercept, but it’s working well for my use case.

PR #343, opened 2025-07-22
feat: add run mode: step

CaiqueMitsuoka:
Hey 👋

This PR introduces a new mode called `step` for `LangChain.Chains.LLMChain.run/2`.

## Context

At Cloudwalk, we are using Langchain to serve 4M users, all with an active assistant.
With that, we have been hitting a few use cases that are not supported by langchain which has made us use a [fork](https://github.com/cloudwalk/langchain/tree/o-3-3-with-stop) with an option to raise an error from tool calls.
This has been hard to keep up with new releases and also contribute back to the project.

## The problem

The existing chain execution modes (`:until_success`, `:while_needs_response`, and the default mode) don't provide the granular control we need for deterministic workflows in regulated environments.
**Our core challenge**: We need to conditionally stop execution based on which tools are called, but the current modes force us to decide the execution strategy upfront, before we know what tools the LLM will choose.
Consider this hipothetical scenario: We have a chain with two tools:

- `wire_transfer(amount, account)` - Executes a money transfer. **Must stop for user confirmation**.
- `user_debts()` - Returns outstanding debts. **Should continue to let LLM explain the results**.

The problem becomes apparent when a user says "Pay my debts":

1. LLM calls user_debts() to check what's owed ✅
2. Tool returns debt information ✅
3. LLM calls wire_transfer() to pay the debts ⚠️
4. We needed to stop here for confirmation, but we can't

### Why existing modes fail:

`:until_success` - Stops after the first successful tool call. Works for `wire_transfer` but fails for `user_debts` (No explanation of the result).
`:while_needs_response` - Continues until LLM provides a final response. Works for `user_debts` but fails for `wire_transfer` (Will explain the tool result, it shouldn't).
Default mode - Behaves like `:until_success`, same limitations.

| Use Case      | `:until_success`     | `:while_needs_response`         | What We Need             |
| ------------- | -------------------- | ------------------------------- | ------------------------ |
| Wire Transfer | ✅ Stops correctly    | ❌ Continues, executes transfer  | Stop after tool call     |
| Check Debts   | ❌ No LLM explanation | ✅ Provides explanation          | Continue after tool call |
| Pay Debts     | ❌ Stops too early    | ❌ Doesn't stop for confirmation | Conditional stopping     |

**The fundamental issue**: We can't know which execution mode we need until we see which tools the LLM decides to call. By then, we're already committed to a mode that might be wrong for the scenario.

What we need is the ability to inspect the chain state after each step and decide whether to continue - giving us the control to implement the deterministic behavior our regulated environment requires.

## The solution

Optionally giving control of the loop. While the loop has served and still is serving us well for some use cases it imposes the aforementioned limitations.
A **stepped execution will give clients control over execution to implement their own loops**. While still being able to call other modes in the middle of the execution.

With this mode we are able to stop on the major chain events: assistant response, tool calls and tool responses.
While still leveraging fallback models and all the data modeling from Langchain.

```elixir
def custom_run(chain, opts \\ []) do
  {:ok, updated_chain} = LLMChain.run(chain, opts ++ [mode: :step])

  if check_some_condition(updated_chain) do
    {:ok, updated_chain}
  else
    custom_run(updated_chain, opts)
  end
end
```

This allows clients to inspect the entire chain state and even alter chain state between executions.
For our use case, this will allow us to achieve the UX we need, and will allow new use cases like extending the chain with dynamic few-shot examples/tools in mid execution for example.

### DIFF
--- a/lib/chains/llm_chain.ex
+++ b/lib/chains/llm_chain.ex
@@ -369,8 +369,8 @@ defmodule LangChain.Chains.LLMChain do
   ## Options
 
   - `:mode` - It defaults to run the chain one time, stopping after receiving a
-    response from the LLM. Supports `:until_success` and
-    `:while_needs_response`.
+    response from the LLM. Supports `:until_success`, `:while_needs_response`,
+    and `:step`.
 
   - `mode: :until_success` - (for non-interactive processing done by the LLM
     where it may repeatedly fail and need to re-try) Repeatedly evaluates a
@@ -391,6 +391,12 @@ defmodule LangChain.Chains.LLMChain do
     are evaluated, the `ToolResult` messages are returned to the LLM giving it
     an opportunity to use the `ToolResult` information in an assistant response
     message. In essence, this mode always gives the LLM the last word.
+
+  - `mode: :step` - (for step-by-step execution control) Executes one step of
+    the chain: makes an LLM call, processes the message, executes any tool
+    calls, and then stops. This allows the caller to inspect the result and
+    decide whether to continue by calling `run` again. Perfect for scenarios
+    where you need to examine each message or tool result before proceeding.
 
   - `with_fallbacks: [...]` - Provide a list of chat models to use as a fallback
     when one fails. This helps a production system remain operational when an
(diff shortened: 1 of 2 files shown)


nallwhy:
This PR is likely to resolve https://github.com/brainlid/langchain/issues/293 also.

brainlid:
This may also be a way to resolve #345.

brainlid (approved):
Looks good! Thanks for the great background on the problem being solved and how this approach helps.

brainlid:
I'm considering adding an option to with something like:

```elixir
LLMChain.run(chain, mode: :step, should_continue?: &check_some_condition/1)
```

It would basically be like the `custom_run` example, but it could do it internally based on the result of the function.

Would that be helpful in your situation as well?

CaiqueMitsuoka:
Thank you @brainlid 😄

It does! And looks super clean. 💪

PR #345, opened 2025-07-24
feat: add support for multiple tools in run_until_tool_used

fortmarek:
Hey 👋 

First of all, thanks a lot for this repo ❤️ 

This PR adds support for adding multiple tools in `LLMChain.run_until_tool_used` in case you want to exit the loop early when either of multiple tools are called.

Our use-case specifically was that we wanted to delete some messages to keep the context window low when a specific tool was called – and then we continue with the execution until a tool signalizing a complete end of the chain is called.

Let me know if that's something you'd consider adding and if so, if you'd implement this differently. I'm new to using `LangChain`, so I wouldn't be surprised if there was a better way to go about this that I'm not aware of 😉

### DIFF
--- a/lib/chains/llm_chain.ex
+++ b/lib/chains/llm_chain.ex
@@ -146,6 +146,22 @@ defmodule LangChain.Chains.LLMChain do
   - The updated chain with all messages and tool calls
   - The specific tool result that matched the requested tool name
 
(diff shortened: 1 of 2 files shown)


brainlid:
This makes sense and I see the value of it. Thanks for the updated docs and tests!
❤️💛💙💜

PR #352, opened 2025-08-04
Prevent crash when ToolResult with string in ChatGoogleAI.for_api/1

nallwhy:
This PR resolves crash in `for_api/1` when `ToolResult` content is a string.

The original bug caused the following error during the process of converting the API request input to json.

```
** (Protocol.UndefinedError) protocol Jason.Encoder not implemented for type LangChain.Message.ContentPart (a struct), Jason.Encoder protocol must always be explicitly implemented.
```


brainlid:
Nice! Thanks @nallwhy!
❤️💛💙💜

PR #356, opened 2025-08-12
 Bedrock OpenAI-compatible API compatibility fix

stevehodgkiss:
Bedrock has support for the gpt-oss models via an OpenAI-compatible API [1] [2].

To test, create a Bedrock API key in the console [3] and configure `ChatOpenAI` with it, a supported gpt-oss model and the correct endpoint:

```elixir
      ChatOpenAI.new(%{
        model: "openai.gpt-oss-120b-1:0",
        api_key: api_key,
        endpoint: "https://bedrock-runtime.us-west-2.amazonaws.com/openai/v1/chat/completions",
      })
```

I had to stop `frequency_penalty` from being sent in order to get past this error: "The 'frequency_penalty' parameter is not supported". This should have no effect on existing code since the parameter is optional and also defaults to 0 (the same as this config) [4]

It'd be nice if support for this new OpenAI-compatible API was extended to the other models on Bedrock 🤞 

[1] https://aws.amazon.com/blogs/aws/openai-open-weight-models-now-available-on-aws/
[2] https://docs.aws.amazon.com/bedrock/latest/userguide/inference-chat-completions.html
[3] https://us-west-2.console.aws.amazon.com/bedrock/home?region=us-west-2#/api-keys?tab=short-term
[4] https://platform.openai.com/docs/api-reference/introduction

### DIFF
--- a/lib/chat_models/chat_open_ai.ex
+++ b/lib/chat_models/chat_open_ai.ex
@@ -233,7 +233,7 @@ defmodule LangChain.ChatModels.ChatOpenAI do
     # Number between -2.0 and 2.0. Positive values penalize new tokens based on
     # their existing frequency in the text so far, decreasing the model's
     # likelihood to repeat the same line verbatim.
-    field :frequency_penalty, :float, default: 0.0
+    field :frequency_penalty, :float, default: nil
 
     # Used when working with a reasoning model like `o1` and newer. This setting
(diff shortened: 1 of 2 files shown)


brainlid:
Awesome! Thanks @stevehodgkiss!
❤️💛💙💜

PR #357, opened 2025-08-15
Add req_config to ChatMode.ChatGoogleAI

nallwhy:
Implemented changes based on the discussion in #351


brainlid:
Thanks! It looks clean and I appreciate trying it out on ChatGoogleAI first. 

Is there a benefit to it being a map then converting it to a keyword list? Or should the `req_config` just be a keyword list?

After looking into it more, I think most things people would want to do are in the config options and creating the %ReqConfig{} style struct probably doesn't make sense here.

nallwhy:
@brainlid 

I initially opted for map because Ecto schemas do not have a native way to represent keyword. This also leaves room for potential serialization in the future. However, switching to `:any` to allow keyword input directly could be a reasonable alternative.

PR #361, opened 2025-08-18
Adds should_continue? optional function to mode step

CaiqueMitsuoka:
As suggested by @brainlid at https://github.com/brainlid/langchain/pull/343.

This adds a new `should_continue?` option for `mode: :step` that provides a cleaner way to manage the execution loop by implementing conditional. Read the mentioned PR for better explanation of the use case.

Added a new use case and update the `mode: :step` use case to highlight when to use it solo.

## Key benefits
  - Internally handles the loop logic for stepped execution
  - Allows fine-grained control of the `run/2` loop when no chain modification is required
  - Streamlines scenarios like max iterations, error thresholds and guardrails.
  
##   Usage
  ```elixir
  should_continue_fn = fn chain ->
    chain.needs_response && length(chain.exchanged_messages) < 10
  end

  {:ok, final_chain} = LLMChain.run(chain, mode: :step, should_continue?: should_continue_fn)
```

### DIFF
--- a/lib/chains/llm_chain.ex
+++ b/lib/chains/llm_chain.ex
@@ -410,9 +410,19 @@ defmodule LangChain.Chains.LLMChain do
 
   - `mode: :step` - (for step-by-step execution control) Executes one step of
     the chain: makes an LLM call, processes the message, executes any tool
-    calls, and then stops. This allows the caller to inspect the result and
-    decide whether to continue by calling `run` again. Perfect for scenarios
-    where you need to examine each message or tool result before proceeding.
+    calls, and then stops. This allows the caller to inspect messages and
+    modify the chain between steps before deciding whether to continue by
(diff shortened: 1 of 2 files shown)


CaiqueMitsuoka:
@brainlid I had a little bit of time in my hand and thought I would take a crack at it, if you don't mind 😄

I have one question: I made it required to pass the `mode: :step` with `should_continue?` since this is more explicit.
I think makes it helps to have the mental model of whats is happening under the hood.
But if you prefer the api just `{:ok, final_chain} = LLMChain.run(chain, should_continue?: should_continue_fn)` I can make a patch, just let me know.

brainlid:
Thanks @CaiqueMitsuoka! I appreciate you taking up the task and finishing it. I agree with you that the `mode: :step` should still be explicitly used. It's clearer and prevents someone from trying to combine `should_continue?` with another mode.

PR #368, opened 2025-09-02
fix: handle LiteLLM responses with null b64_json in OpenAIImage

fbettag:
## Summary

Fixes compatibility issue with LiteLLM when using OpenAI image generation. LiteLLM returns responses where `b64_json` is `null` when providing URLs, which was causing "content: can't be blank" errors.

## Changes

- Added guard clause `when is_binary(base64_raw_content)` to ensure the b64_json pattern only matches when there's actual base64 content, not `null`
- Added test case to verify LiteLLM-style responses work correctly
- Maintains backward compatibility with existing behavior

## Test Plan

- [x] Added test case that verifies handling of LiteLLM response format
- [x] Existing tests continue to pass for standard OpenAI responses
- [x] Fix resolves the "content: can't be blank" error when b64_json is null but url is present

## Related Issues

This addresses compatibility with LiteLLM proxy responses that set `b64_json: null` while providing image URLs.

### DIFF
--- a/lib/chat_models/chat_open_ai.ex
+++ b/lib/chat_models/chat_open_ai.ex
@@ -1008,8 +1008,10 @@ defmodule LangChain.ChatModels.ChatOpenAI do
   # Delta message tool call
   def do_process_response(
         model,
-        %{"delta" => delta_body, "finish_reason" => finish, "index" => index} = _msg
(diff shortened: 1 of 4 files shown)


brainlid:
Thanks @fbettag! I appreciate the improved compatibility.
❤️💛💙💜

PR #371, opened 2025-09-06
Add `parallel_tool_calls` option to `ChatOpenAI` model

martosaur:
I couldn't find an escape hatch that would allow me to pass `parallel_tool_calls: false` to the OpenAI API endpoint, so naturally here's a PR adding an explicit support for it!


brainlid:
Thanks @martosaur! It's nicely done simple addition that improved the developer experience. Thanks for the included test coverage too.
❤️💛💙💜

PR #372, opened 2025-09-07
Add optional AWS session token handling in BedrockHelpers

quangngd:
This change allows for improved handling of AWS credentials, accommodating scenarios where a session token may be required (e.g. when use credentials given by AWS Sts)

- Introduced a new private function `maybe_add_token/1` to conditionally add the AWS session token to the credentials if it is available in the application environment.
- Updated the credentials retrieval logic to incorporate this new functionality, enhancing the flexibility of AWS credential management.


brainlid:
Hi @quangngd! Can you please update the `LangChain.Utils.BedrockConfig` module documentation so people will know this config option exists? Also, a brief explanation of when and why this config is needed would be helpful for other users. The goal is to expose the supported config name and explain how it's used. 

I'd also love to see a test covering it. :slightly_smiling_face:

brainlid:
Thanks for the updates!
❤️💛💙💜

PR #373, opened 2025-09-08
fix(ChatGoogleAI): Handle cumulative token usage

mweidner037:
When streaming with ChatGoogleAI, empirically, Google returns the *cumulative* token usage with each delta - unlike other models that return only the incremental token usage. Thus we need to avoid adding together the delta's token usages, instead just taking the last one.

I noticed this by comparing usage stats to those reported in [Helicone](https://www.helicone.ai/) for the same requests. With this PR, they match exactly.


brainlid:
Thanks @mweidner037! I like it. :slightly_smiling_face: 
❤️💛💙💜

PR #375, opened 2025-09-08
feat(ChatGoogleAI): Full thinking config

mweidner037:
In https://github.com/brainlid/langchain/pull/354, the option to configure thinking for ChatGoogleAI was added. However, you can only control the thinkingBudget; includeThoughts defaults to true whenever the budget is nonzero.

This PR allows you to configure both options as a map, specifically, an arbitrary [ThinkingConfig](https://ai.google.dev/api/generate-content#ThinkingConfig). That is preferable for our app (we don't need thought summaries), and it is more flexible in case Google adds further thinking options in the future.


brainlid:
Thanks @mweidner037, that change makes sense.
❤️💛💙💜

PR #379, opened 2025-09-15
Support verbosity parameter for ChatOpenAI

rohan-b99:
GPT-5 has a new `verbosity` parameter: https://platform.openai.com/docs/api-reference/chat/create#chat-create-verbosity

This PR adds support for optionally passing it in (defaults to `nil`)


brainlid:
Thanks @rohan-b99! Much appreciated!
❤️💛💙💜

PR #384, opened 2025-09-25
Use moduledoc instead of doc for LLMChain documentation

xxdavid:
Using `@doc` was probably a typo.


brainlid:
Good catch! Thanks!
❤️💛💙💜

PR #391, opened 2025-10-08
OpenAI responses API improvements

arjan:
* Add usage tracking for non-streaming responses
* Process file_search_call content part when `file_search` is added as native tool


brainlid:
Thanks @arjan!
❤️💛💙💜

arjan:
You're welcome!

PR #392, opened 2025-10-14
feat: Consistently include `original` in "Unexpected response" errors

mweidner037:
We're encountering these while using ChatGoogleAI but with `original: nil`. The PR adds `original` in that case, as well as in the other Chat... classes at an analogous place.

Note: There is a difference in the error messages returned in the "Unexpected response" case for streaming vs non-streaming, consistently across chat models. If you like, I can make them consistent with each other, in either direction.
```elixir
# do_api_request(...{stream: false}...):
      other ->
        Logger.error("Unexpected and unhandled API response! #{inspect(other)}")
        other

# do_api_request(...{stream: true}...):
      other ->
        Logger.error(
          "Unhandled and unexpected response from streamed post call. #{inspect(other)}"
        )

        {:error,
         LangChainError.exception(
           type: "unexpected_response",
           message: "Unexpected response",
           original: other
         )}
```

### DIFF
--- a/lib/chat_models/chat_anthropic.ex
+++ b/lib/chat_models/chat_anthropic.ex
@@ -613,11 +613,8 @@ defmodule LangChain.ChatModels.ChatAnthropic do
         error
 
       other ->
-        message = "Unexpected and unhandled API response! #{inspect(other)}"
-        Logger.error(message)
-
-        {:error,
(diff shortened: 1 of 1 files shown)


brainlid:
Thanks @mweidner037! Sorry it took me so long to follow-up on this. I appreciate your contribution and work though!
❤️💛💙💜

PR #393, opened 2025-10-16
Update gettext dependency version to 1.0

bijanbwb:
### Context

Tiny PR to update the `gettext` dependency from `0.26` to `1.0`.

It sounds like a big update to `1.0`, but there are no breaking changes. Here is an excerpt from the CHANGELOG:

> There are _very few changes_ from the latest 0.26 release, and none of them are breaking.

### References

https://github.com/elixir-gettext/gettext/blob/v1.0.0/CHANGELOG.md#v100


brainlid:
Thanks @bijanbwb! And thanks for the link to the CHANGELOG.
❤️💛💙💜

PR #394, opened 2025-10-24
Add DeepSeek chat model integration

gilbertwong96:
- Add ChatDeepSeek module for DeepSeek API integration
- Add comprehensive test suites
- Update documentation and package configuration
- Add `DEEPSEEK_API_KEY` to environment template


brainlid:
Hi @gilbertwong96! Thank you for the DeepSeek contribution!

I'm wondering if there's a way to create the ChatDeepSeek module with the default config you want (API endpoint, model name, etc) while reusing the ChatOpenAI implementation? 

If it's supposed to be OpenAI compatible, then I'd rather not duplicate the whole module and tests because it becomes a larger support burden as changes come later.

If there are problems with supporting the ChatOpenAI code that you had to fix for DeepSeek support, then I'd love to see those changes put into the ChatOpenAI module, viewing it as something was wrong for cross-service compatibility.

You are closer to this and what is needed for DeepSeek support. Would that approach work here?

gilbertwong96:
Hi, @brainlid, yes, I have tried ChatOpenAI implementation before and indeed I encountered some issues:


```
12:38:16.399 [error] Error during chat call. Reason: %LangChain.LangChainError{type: nil, message: "Failed to deserialize the JSON body into the target type: messages[1]: invalid type: sequence, expected a string at line 1 column 339", original: %{"error" => %{"code" => "invalid_request_error", "message" => "Failed to deserialize the JSON body into the target type: messages[1]: invalid type: sequence, expected a string at line 1 column 339", "param" => nil, "type" => "invalid_request_error"}}}
```

The error above occurs because DeepSeek still uses the legacy API schema OpenAI used before.

From the [DeepSeek API Docs](https://api-docs.deepseek.com/api/create-chat-completion), we can see that the `content` field is required using `String` type data.

image

However, based on the [OpenAI Docs](https://platform.openai.com/docs/api-reference/responses/create), we can see that the `content` field looks as follows:

```json
[
  {
    "type": "input_text",
    "text": "what is in this file?"
  },
  {
    "type": "input_file",
    "file_url": "https://www.berkshirehathaway.com/letters/2024ltr.pdf"
  }
]
```

I think it's not a good practice to make `ChatOpenAI` implementation to support the API schema of DeepSeek. As a result, I make a new implementation of `ChatDeepSeek` in this PR.

brainlid:
Thanks @gilbertwong96 for the background and explanation! I appreciate it. And thanks for the contribution!
❤️💛💙💜

PR #395, opened 2025-10-26
Open AI Responses API: Add support for file_url with link to file

reetou:
Noticed that [this](https://platform.openai.com/docs/guides/pdf-files#file-urls) part of the API doesn't work and crashes if you add a `ContentPart.file_url!` part with `LangChain.ChatModels.ChatOpenAIResponses`. Tested locally to confirm that it works, let me know if you want me to add any more tests

```js
const response = await client.responses.create({
    model: "gpt-5",
    input: [
        {
            role: "user",
            content: [
                {
                    type: "input_text",
                    text: "Analyze the letter and provide a summary of the key points.",
                },
                {
                    type: "input_file", // This one should have `file_url`, not content
                    file_url: "https://www.berkshirehathaway.com/letters/2024ltr.pdf",
                },
            ],
        },
    ],
});
```

### DIFF
--- a/lib/chat_models/chat_open_ai_responses.ex
+++ b/lib/chat_models/chat_open_ai_responses.ex
@@ -541,6 +541,16 @@ defmodule LangChain.ChatModels.ChatOpenAIResponses do
 
   def content_part_for_api(
         %ChatOpenAIResponses{} = _model,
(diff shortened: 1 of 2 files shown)


brainlid:
Thanks @reetou! Yay!
❤️💛💙💜

PR #396, opened 2025-10-27
Add dialyxir for type checking

gilbertwong96:
- Update the dependency: [ex_doc](https://hex.pm/packages/ex_doc)
- Add the dependency: [dialyxir](https://hex.pm/packages/dialyxir)
- Resolve dialyzer type inference issues across codebase


brainlid:
Thanks @gilbertwong96. My current plan is to not focus on Dialyzer at this point. I've been using primarily for documentation only. I'm planning to fully embrace the new Elixir type system syntax when it comes out.

PR #397, opened 2025-10-29
add Claude NativeTool support

KristerV:
i was trying to use the Claude web search tool, but NativeTool crashed on me. So I made a few changes.

This solution works, but the type/name situation is a tad confusing. Claude wants both keys to be present and one can be derived from the other. Docs here: https://anthropic.mintlify.app/en/docs/agents-and-tools/tool-use/web-search-tool

Anyway ready to be merged imo, but open to suggestions also. I used this in prod, but now am actually switching to Gemini, because it reads handwriting the best :D

### DIFF
--- a/lib/chat_models/chat_anthropic.ex
+++ b/lib/chat_models/chat_anthropic.ex
@@ -200,6 +200,7 @@ defmodule LangChain.ChatModels.ChatAnthropic do
   alias LangChain.TokenUsage
   alias LangChain.Function
   alias LangChain.FunctionParam
+  alias LangChain.NativeTool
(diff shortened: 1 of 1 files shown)


xxdavid:
Hi, thanks for the PR. Works great!

I think it would also be good to handle the [web fetch](https://platform.claude.com/docs/en/agents-and-tools/tool-use/web-fetch-tool) results. Like this: https://github.com/xxdavid/langchain/commit/814c493071cf4443d3007e291b7a411528e46306. What do you think?

KristerV:
personally not using Claude atm. but yeah, good idea.

PR #398, opened 2025-10-29
Added an :cache_messages option for ChatAnthropic, can improve cache utilization.

montebrown:
- for multi-turn conversations, Anthropic recommends placing a cache breakpoint on the last user message to achieve high cache hit rates.
- The `:cache_messages` option automates this pattern by automatically adding `cache_control` to the last user message's last text content part.
- This is particularly effective for conversations because as new turns are added, the cache breakpoint automatically moves to the newest message, maximizing cache utilization across the conversation history.
- Supports optional TTL configuration ("5m" or "1h")

Usage Example 
model = ChatAnthropic.new!(%{ model: "claude-3-5-sonnet-20241022", cache_messages: %{enabled: true} })

See:
https://github.com/anthropics/claude-cookbooks/blob/main/misc/prompt_caching.ipynb

### DIFF
--- a/lib/chat_models/chat_anthropic.ex
+++ b/lib/chat_models/chat_anthropic.ex
@@ -193,64 +193,182 @@ defmodule LangChain.ChatModels.ChatAnthropic do
 
   ### Automatic Message Caching for Multi-Turn Conversations
 
-  For multi-turn conversations, Anthropic recommends placing a cache breakpoint on the last user message
-  to achieve high cache hit rates. The `:cache_messages` option automates this pattern by automatically
-  adding `cache_control` to the last user message's last text content part.
+  The `:cache_messages` option automates cache breakpoint placement for multi-turn conversations by
+  adding `cache_control` to the last N user messages.
 
-  This is particularly effective for conversations because as new turns are added, the cache breakpoint
-  automatically moves to the newest message, maximizing cache utilization across the conversation history.
+  #### When to Use This Feature
 
-  With :cache_messages enabled:
-  - The cache breakpoint is only added to the **last user message** in the conversation
-  - It's added to the **last text ContentPart** within that message (skips tool_result types)
-  - This works alongside system message caching and tool caching
+  **Good fit:**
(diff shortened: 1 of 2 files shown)


brainlid:
I assume you've run this against a live environment and verified the cache reads work as you expect?

I like the approach and it would make it easier for users to benefit from Anthropic caching without having to manage the cache points themselves. That's what I've been doing, is managing it in the application. But this would be handy. :slightly_smiling_face:

montebrown:
I have! I've tested it across a range of prompts - locally and it's been in production for a little while now. In my local testing I saw behavior in-line with the python example from Anthropic.

I've seen a jump in average cache utilization (>30% with utilization now spiking over 90%) and significantly faster multi-turn conversations. 

In testing I found a couple gotchas that can lead to lower than expected utilization. I'll add add a section or two to the docs to include this as well as some info on monitoring utilization and what to expect.

montebrown:
Ok! So that took quite a bit longer than expected.

In reviewing the numbers I noticed steep utilization degradation for multi-turn conversations with heavy tool use. 

It turns out that the cookbook example is overly simplistic and doesn't address multi-turn dynamics. The documentation wasn't detailed enough so I consulted Claude on the behavior and verified the answers. It doesn't address that with the simple approach, tool use progressively pushes down utilization since it interferes with prefix-matching. There are a lot of details.

Some degradation of utilization is unavoidable with long conversations and only 4 breakpoints, but I tested this solution and I see better numbers for long conversations with tool use. Utilization will still eventually drop and with large numbers of successive tool calls I managed to get it down to 34%, but it got there more gradually. Before most prompts degraded from 99+% immediately, now many maintain that level past 3 turns. 

Hopefully the my additions to the docs make it more clear how to get good cache utilization without being too verbose.

brainlid:
@montebrown Great! One of the approaches I've used in my application was to only add a cache point to certain tool results that returned large data. Meaning I didn't keep moving the cache points until another large operation took place. And those are fairly specific to my application and use case.

I like the addition of the docs. It'll be helpful.
❤️💛💙💜

PR #399, opened 2025-10-29
Loosen the gettext dependency

montebrown:
- it's an unusual situation: gettext published a major version with no breaking changes.
- since it was a major version upgrade this is incompatible with libraries that still require 0.26.x.
- with no breaking changes between the versions, allowing both comes with little risk and fewer apps run into a dependency mess.

https://github.com/elixir-gettext/gettext/blob/v1.0.0/CHANGELOG.md#v100


brainlid:
Nice! I like your approach. Thanks @montebrown!
❤️💛💙💜

PR #403, opened 2025-11-12
Add support for Anthropic API PDF reading to the ChatAnthropic model.

jadengis:
Anthropic has added support for parsing PDFs to the API similar to OpenAI.

This PR adds support for that feature to the ChatAntropic Model.

https://docs.claude.com/en/docs/build-with-claude/pdf-support#process-pdfs-with-claude


brainlid:
Thanks @jadengis! Great addition!
❤️💛💙💜

PR #404, opened 2025-11-16
feat: Add support for `:file_url` to ChatAnthropic too

jadengis:
I kind of willfully ignored this piece in the last PR I posted and realizing that URLs are probably the best way to let you LLM provider stream content, decided it would be best if I closed that gap.

https://docs.claude.com/en/docs/build-with-claude/pdf-support#option-1%3A-url-based-pdf-document


brainlid:
Thanks @jadengis!
❤️💛💙💜

PR #407, opened 2025-11-18
Support reasoning_content of deepseek model

gilbertwong96:
Unlike ChatGPT, DeepSeek model only supports one choice and there's only one index `0` in DeepSeek response. However, I wish I could get the thinking text from API response when I evaluate `LLMChain.run()`. As a result, I utilize the index `0` to place the thinking message and the index `1` to place the text message.


brainlid:
Thanks @gilbertwong96! That's an interesting adaptation to getting the thinking contents. But it works!
❤️💛💙💜

kayuapi:
Hi, there is no need to modify the index right? this is because the thinking content and the content has been put into %MessasgeDelta{content: **here**}

PR #408, opened 2025-11-18
Add req_opts to ChatAnthropic

stevehodgkiss:
Hey @brainlid 👋 

Following on from the discussion & PRs in https://github.com/brainlid/langchain/issues/347, this PR adds the same functionality to ChatAnthropic.

The first commit follows the same approach as https://github.com/brainlid/langchain/pull/357, `req_config` using a map (which would be serializable).

The second commit switches to `req_opts` with a virtual keyword list. I have a slight preference for this approach since req takes a keyword list, but we may need to implement a custom `Ecto.Type` to make it serializable (if this is needed). Thoughts?


brainlid:
Nice @stevehodgkiss! I like it! Thanks!
❤️💛💙💜

PR #410, opened 2025-11-24
Allow multi-part tool responses.

montebrown:
- previously if a tool response was mult-part it would return an error.
- tools that return images often return a 2-part message as in the test case.


brainlid:
Thanks @montebrown! I haven't run into that situation before. Much appreciated!
❤️💛💙💜

PR #416, opened 2025-12-12
Add verbose_api field to ChatPerplexity and ChatMistralAI

arjan:
This fixes an issue where handle_stream_fn in Utils would fail when accessing model.verbose_api on models that didn't define this field.

This fixes using Mistral for me using the streaming API. Have not checked Perplexity TBH but the change is small.


brainlid:
Thanks @arjan!
❤️💛💙💜

PR #417, opened 2025-12-12
Fix Mistral AI tool result content formatting

arjan:
ToolResult content can be a list of ContentParts, but Mistral API expects tool result content to be a string. The for_api function now properly converts ContentParts to strings using ContentPart.parts_to_string/1.


brainlid:
Is this still an issue in light of the other merged PR #433? I realize I was merging from most recent going back to older and accepted the other one first. Just let me know! If there's still a fix needed for what you're doing, feel free to update this PR based on the current code and we'll take it from there.

arjan:
Yes I guess this can be closed now, thanks

PR #419, opened 2025-12-15
Add new reasoning effort values to ChatOpenAIResponses

xxdavid:
GPT 5.1 also allows `none`, GPT 5.2 also allows `xhigh`.

Note that GPT 5.2 currently fails with ChatOpenAIResponses when using reasoning effort other than `none`:

    %LangChain.LangChainError{
        type: nil,
        message: "Unsupported parameter: 'top_p' is not supported with this model.",
        original: nil
    }

but I think that should be fixed in a separate commit/PR.

See: https://platform.openai.com/docs/guides/latest-model#gpt-5-2-parameter-compatibility

### DIFF
--- a/lib/chat_models/reasoning_options.ex
+++ b/lib/chat_models/reasoning_options.ex
@@ -6,7 +6,8 @@ defmodule LangChain.ChatModels.ReasoningOptions do
 
   ## Fields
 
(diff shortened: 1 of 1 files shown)


brainlid:
Thanks @xxdavid! Gotta love reasoning! :thinking: 
❤️💛💙💜

PR #420, opened 2025-12-15
fix ChatGrok tool call arguments and message flattening

KristerV:
not sure if grok isn't used at all or is this a versioning issue. but i decided to use it and couldn't. seems that the format Grok replies in has some differences in how langchain expects them. so i made the necessary changes so that Grok works.


brainlid:
Thanks @KristerV!
❤️💛💙💜

brainlid:
Hey @KristerV, I had to revert it in #445.

It broke a number of tests and I'm trying to finalize a new release, so it wasn't quite ready.

PR #421, opened 2025-12-15
Add support for OpenAI reasoning/thinking events in ChatOpenAIResponses

arjan:
Thinking content is now properly streamed and accumulated separately from regular text content using output_index tracking. This prevents content type mismatches when OpenAI returns both reasoning and message output items in the same response.


brainlid:
Thanks @arjan! Looks good! :grin: Sorry for the long delay. I appreciate your contributions.
❤️💛💙💜

PR #423, opened 2025-12-18
Add thinking config to vertex ai

mattmatters:
Google is getting better at not diverging specs between both their ai offerings. This essentially just ports the logic from Google AI to Vertex.

I am currently running this in production as well.


brainlid:
Thanks @mattmatters! 
❤️💛💙💜

PR #425, opened 2025-12-20
Add support for OpenAI Response API Stateful context

cjimison:
In OpenAI Response, you can use Stateful Context, which keeps track of your chat history so you don’t have to upload it again and again. Langchain is aiming for a completely stateless approach (as mentioned in the code), but by using this system, it can save you both time and money—less bandwidth and fewer tokens, according to OpenAI.

These changes enable the following use case:

```elixir
# Start a converstation
def start_story() do
    {:ok, chain} =
          # NOTE:  Enable OpenAI "store" system.  this is to maintain backwards compatibility with previous version
          #        of langchain and for folks that don't want to use the store system.  Default is still false
           %{llm: ChatOpenAIResponses.new!(%{model: "gpt-5-nano", store: true})}
           |> LLMChain.new!()
           |> LLMChain.add_message(Message.new_system!("You are a sci-fi author and expert in space travel."))
           |> LLMChain.add_message(Message.new_user!("Tell me a story!"))
           |> LLMChain.run()

    ContentPart.content_to_string(chain.last_message.content)
    |> Logger.info()
    chain.last_message.metadata.response_id
end
# Now store the response_id in the DB or message it out to another node in the cluster, etc.  I do this via using Oban.

def continue_story(response_id) do
    {:ok, chain} =
           # NOTE:  Pass in the previous response id to pick up where we last left off without having send up full history
           %{llm: ChatOpenAIResponses.new!(%{model: "gpt-5-nano", previous_response_id: response_id})}
           |> LLMChain.new!()
           |> LLMChain.add_message(Message.new_user!("What happens next?"))
           |> LLMChain.run()
    ContentPart.content_to_string(chain.last_message.content)
    |> Logger.info()
    chain.last_message.metadata.response_id
end

def finish_story(response_id) do
    {:ok, chain} =
           # NOTE:  Pass in the previous response id to pick up where we last left off without having send up full history
           %{llm: ChatOpenAIResponses.new!(%{model: "gpt-5-nano", previous_response_id: response_id})}
           |> LLMChain.new!()
           |> LLMChain.add_message(Message.new_user!("Finish the tale!"))
           |> LLMChain.run()
    
    ContentPart.content_to_string(chain.last_message.content)
    |> Logger.info()
    chain.last_message.metadata.response_id
end

# This function is just a simple example but you can image each of this function calls happening on another server
# or after a Kubernetes pod restart, etc.  The advantage here is I don't have go fetch all the previous change
# messages,etc.  I just need to keep the response_id to pickup where I last left off.
def create_story() do
    rsp_id = start_story()
    # You can now store and later fetch that rsp_id to pickup the chat
    rsp_id = continue_story(rsp_id)
    rsp_id = continue_story(rsp_id)
    rsp_id = finish_story(rsp_id)
end
```

let me know what folks think.  I have been using this on my system and things look good so far.  I don't think I broke any of the other models like Claude, etc.

### DIFF
--- a/lib/chat_models/chat_open_ai_responses.ex
+++ b/lib/chat_models/chat_open_ai_responses.ex
@@ -216,11 +216,11 @@ defmodule LangChain.ChatModels.ChatOpenAIResponses do
     field :max_output_tokens, :integer, default: nil
     # omit metadata because chat_open_ai also omits it
     # omit parallel_tool_calls because chat_open_ai also omits it
-    # omit previous_response_id becasue langchain assumes statelessness
+    field :previous_response_id, :string, default: nil
     # Reasoning options for gpt-5 and o-series models
     embeds_one(:reasoning, ReasoningOptions)
     # omit service_tier because chat_open_ai also omits it
-    # omit store, but set it explicitly to false later to keep statelessness. the API will default true unless we set it
+    field :store, :boolean, default: false
     field :stream, :boolean, default: false
     field :temperature, :float, default: nil
     field :json_response, :boolean, default: false
@@ -252,6 +252,8 @@ defmodule LangChain.ChatModels.ChatOpenAIResponses do
     :model,
     :include,
     :max_output_tokens,
+    :previous_response_id,
+    :store,
     :stream,
     :temperature,
     :json_response,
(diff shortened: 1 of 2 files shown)


brainlid:
Wow @cjimison, this is great!
❤️💛💙💜

My only additional request would be to expand the ChatOpenAI's moduledoc with a recommended usage like in your PR description.

It's an interesting counterpoint to Anthropic's caching.

PR #426, opened 2025-12-21
refactor: make chain modes extensible via behaviour pattern

nbw:
Extracts hardcoded modes from single file into separate modules that each implement a common _Mode_ Behaviour, allowing users to define custom modes without modifying the library.

## Comments 

I'm not expecting this PR to get merged necessarily, but I wanted to propose a possible way of making run Modes more approachable with a Behaviour pattern. 

I found myself wanting to write my own "mode" to have control over the restart strategy, but some functions like `do_run` are private and unaccessible. So I'd end up using the :step mode and then copying some of the private functions internals. Perhaps that's the intention of the author, in which case feel free to close this PR. By having a module that defines the Mode behaviour, you can pass it in as an option and the higher level api of calling LLMChain.run(...) stays the same in your code.

### DIFF
--- a/lib/chains/llm_chain.ex
+++ b/lib/chains/llm_chain.ex
@@ -190,6 +190,7 @@ defmodule LangChain.Chains.LLMChain do
   alias LangChain.Chains.ChainCallbacks
   alias LangChain.PromptTemplate
   alias __MODULE__
+  alias __MODULE__.Modes
   alias LangChain.Message
   alias LangChain.Message.ContentPart
   alias LangChain.Message.ToolCall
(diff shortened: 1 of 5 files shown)


brainlid:
Hi @nbw! What type of custom behaviors are you wanting to implement?

brainlid:
Hi @nbw! With the new Sagents library, it adds support for HITL (Human in the Loop) which required changes to the run mode.

This is still an interesting idea and I'm still just thinking about it.

I was less involved on this project while I was building out the sagents-ai set of projects. I'm bringing focus back to here now though.

brainlid:
Hey @nbw! Thanks for the design suggestion! I used that module extraction idea but implemented it differently and more broadly. It's in #469.

nbw:
sounds good @brainlid!

Issue #405, opened 2025-11-17
build error with chat_google_ai


darthxvaderxd:
So I am trying to build the latest and am running into an error. Maybe I am missing something, because I am not seeing other people report the same issue and this code has been here for 3 months. However, here is what I am seeing.

```
==> langchain
Compiling 1 file (.ex)
     error: undefined function get_in/1 (expected LangChain.ChatModels.ChatGoogleAI to define such a function or for it to be imported, but none are available)
     │
 600 │               content_type = get_in(message_delta.content.type)
     │                              ^^^^^^
     │
     └─ (langchain 0.4.0) lib/chat_models/chat_google_ai.ex:600:30: LangChain.ChatModels.ChatGoogleAI.do_api_request/3

     error: undefined function get_in/1 (expected LangChain.ChatModels.ChatGoogleAI to define such a function or for it to be imported, but none are available)
     │
 599 │               last_content_type = get_in(last_message_delta.content.type)
     │                                   ^^^^^^
     │
     └─ (langchain 0.4.0) lib/chat_models/chat_google_ai.ex:599:35: LangChain.ChatModels.ChatGoogleAI.do_api_request/3
```

https://github.com/brainlid/langchain/blob/main/lib/chat_models/chat_google_ai.ex#L599C1-L600C64

As far as I can tell get_in needs two parameters. I believe this should be something like the following.

```
last_content_type = get_in(last_message_delta, [:content, :type])
content_type = get_in(message_delta, [:content, :type])
```

darthxvaderxd:
Just answered my own question my apologies: https://hexdocs.pm/elixir/Kernel.html#get_in/1

nbw:
for those wondering what's up, it's a macro that requires elixir v 1.17.0 or later.

PR #427, opened 2025-12-23
use elixir 1.17

nbw:
the ChatGoogleAI module uses the `get_in` macro which is only available in Elixir 1.17 onwards. Using elixir 1.16 throws a compile error.

## Context

I ran into [this issue](https://github.com/brainlid/langchain/issues/405) at compile time since I was using 1.16.


brainlid:
Thanks @nbw! Appreciated!
❤️💛💙💜

PR #428, opened 2025-12-23
Don't include top_p for gpt-5.2+ in ChatOpenAIResponses

montebrown:
- gpt-5.2 and newer models do not support the top_p parameter and return an error when it is included. Earlier models accept top_p.
- Now the top_p parameter is excluded from API requests for unsupported models while preserving the existing behavior for older models.


brainlid:
Thanks @montebrown!
❤️💛💙💜

PR #429, opened 2025-12-27
fix "Support reasoning_content of deepseek model" introducing UI bug …

kayuapi:
…for deepseek-chat model


brainlid:
Thanks @kayuapi! :bow: 
❤️💛💙💜

PR #430, opened 2025-12-31
Handle thought content parts from Google Vertex

mattmatters:
This is again just copying google ai functionality.

The next move is to extract a lot of common stuff from here. I'm thinking we just delegate a lot of functions to `LangChain.ChatModels.ChatGoogleAI` and then override the couple differences/divergences.

The main difference I'm seeing is both are handling file uploads differently (in our code) which may require a deprecation.


brainlid:
Thanks @mattmatters!
❤️💛💙💜

PR #432, opened 2026-01-07
feat(ChatMistralAI): add parallel tool calling, verbose API, add missing converter for tool call result content

xu-chris:
This PR solves three things when using Mistral:
1. Tool call results might be a list. The current implementation wasn't able to handle that
2. While doing so, I have realized that Mistral calls tools in parallel. Mistral supports `parallel_tool_calls` - like OpenAI does
3. `verbose_api` was a required addition for easier bug fixing

I did use Claude Opus 4.5 for creating the changes. I tried to align them as close as possible to the `ChatOpenAI` module.

Hope this PR meets your quality standard. 🙏

### DIFF
--- a/lib/chat_models/chat_mistral_ai.ex
+++ b/lib/chat_models/chat_mistral_ai.ex
@@ -56,6 +56,9 @@ defmodule LangChain.ChatModels.ChatMistralAI do
     # For choosing a specific tool call (like forcing a function execution).
     field :tool_choice, :map
 
+    # Whether to allow the model to make parallel tool calls. `true` to reflect current default of API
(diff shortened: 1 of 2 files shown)


xu-chris:
I've realized that this PR might duplicate solutions of @arjan / PR #417 . 
@brainlid Do you need a hand to check and merge the PRs?

arjan:
Yep I created #416, #417 and #418 for these, they overlap indeed..

xu-chris:
Closed because of overlapping PRs.
Edit: Opened #433 for handling parallel tool calls

PR #433, opened 2026-01-07
feat(ChatMistralAI): add support for parallel tool calls 

xu-chris:
Adding the support for parallel tool calls. 
Tool calls can be therefore result in a list. This PR is handling it as well


brainlid:
Thanks @xu-chris! Nice addition and I appreciate the test coverage.
❤️💛💙💜

PR #435, opened 2026-01-13
fix(ChatMistralAI): missing error handling and fallback mechanism on server outages (#434)

### LINKED ISSUE #434: ChatMistralAI: no support for `unreachable_backend`
Whenever the Mistral service is not available due to an [outage](https://status.mistral.ai), LangChain fails to fallback to the fallback models since the error is not mapped correctly.

Log entry received:
```
09:50:46.965 [error] Received error from Mistral API: "Internal server error"
ERROR: %LangChain.LangChainError{
  type: "unreachable_backend",
  message: "Internal server error",
  original: %{
    "code" => "1100",
    "message" => "Internal server error",
    "object" => "error",
    "param" => nil,
    "type" => "unreachable_backend"
  }
}
09:50:47.010 [error] Error during chat call. Reason: %LangChain.LangChainError{type: "unreachable_backend", message: "Internal server error", original: %{"code" => "1100", "message" => "Internal server error", "object" => "error", "param" => nil, "type" => "unreachable_backend"}}
```

The error seems to be already parsed by LangChain correctly but it fails to handle the error correctly with `retry_on_fallback?`.

The fix seems to be rather simple:
```
def retry_on_fallback?(%LangChainError{type: "unreachable_backend"}), do: true
def retry_on_fallback?(%LangChainError{type: "server_error"}), do: true
def retry_on_fallback?(%LangChainError{type: "service_unavailable"}), do: true
```

Will add a PR later. This ticket is to track this issue.

xu-chris:
Closes #434

### DIFF
--- a/lib/chat_models/chat_mistral_ai.ex
+++ b/lib/chat_models/chat_mistral_ai.ex
@@ -719,6 +719,9 @@ defmodule LangChain.ChatModels.ChatMistralAI do
   def retry_on_fallback?(%LangChainError{type: "rate_limit_exceeded"}), do: true
   def retry_on_fallback?(%LangChainError{type: "timeout"}), do: true
   def retry_on_fallback?(%LangChainError{type: "too_many_requests"}), do: true
+  def retry_on_fallback?(%LangChainError{type: "unreachable_backend"}), do: true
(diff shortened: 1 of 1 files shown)


brainlid:
Thanks @xu-chris!
❤️💛💙💜

Issue #443, opened 2026-01-24
Chain sometimes hangs when unable to decode JSON


sgiath:
I have noticed, that occasionally my subagents get stuck in a loop where they print this logs over and over:

```
  [warning] Error applying delta message. Reason: "tool_calls: arguments: invalid json"
  [warning] Error applying delta message. Reason: "tool_calls: arguments: invalid json"
  [warning] Error applying delta message. Reason: "tool_calls: arguments: invalid json"
  [warning] Error applying delta message. Reason: "tool_calls: arguments: invalid json"
  [warning] Error applying delta message. Reason: "tool_calls: arguments: invalid json"
  [warning] Error applying delta message. Reason: "tool_calls: arguments: invalid json"
  [warning] Error applying delta message. Reason: "tool_calls: arguments: invalid json"
```

I have traced it down to this place:
https://github.com/brainlid/langchain/blob/v0.4.1/lib/chains/llm_chain.ex#L1074-L1078

And the error seems to originate here:
https://github.com/brainlid/langchain/blob/v0.4.1/lib/message/tool_call.ex#L119

I am sure this is an error on the LLM side sending malformed response but I don't think it should be stuck in a loop. It eventually reach some max number of iterations and ends itself but I would expect it to end immediately and maybe print some more information why it failed.

I am using `LLMChian` with `LangChain.ChatModels.ChatOpenAI` on OpenRouter, I experienced it with different models but currently I am on `moonshotai/kimi-k2-0905`

I have created PR with test that exposes the behavior I am describing, hopefully it is clear what I mean. I have also included `debug.log` file with the real scenario where that happened (I have extended what is actually logged in the library to capture this).

brainlid:
Thanks @sgiath! This was fixed in #449 but this issues wasn't closed. I added some test coverage to verify it. This also closes the PR #444.

PR #444, opened 2026-01-24
Infinite loop when LLM sends malformed JSON

sgiath:
Added test that exposes the behavior described in issue https://github.com/brainlid/langchain/issues/443


brainlid:
Closed as fixed

PR #448, opened 2026-01-27
fix(ChatMistralAI): Handle broken tool names that kill LLMChain

xu-chris:
## Summary

Fixes cases when Mistral returns malformed tool call responses that can break the LLMChain execution.

This PR adds defensive handling for:
- Tool calls with corrupted or empty names
- Malformed tool call responses that would otherwise cause message ordering errors

## Changes

- Added validation for tool call names in `ChatMistralAI`
- Added tests for malformed tool call scenarios

## Test plan

- [x] All existing tests pass
- [x] New tests added for malformed tool call handling

### DIFF
--- a/lib/chat_models/chat_mistral_ai.ex
+++ b/lib/chat_models/chat_mistral_ai.ex
@@ -599,40 +599,58 @@ defmodule LangChain.ChatModels.ChatMistralAI do
           nil
       end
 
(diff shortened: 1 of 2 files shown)


brainlid:
Thanks @xu-chris!
❤️💛💙💜

PR #449, opened 2026-01-27
feat: Verify function parameters before executing, reject faulty message deltas

xu-chris:
## Summary

This PR hardens the LLMChain by auto-rejecting malformed tool calls before execution, preventing chain failures with less reliable LLMs like Mistral.

A tool call gets rejected when:
- The tool name is in content or is malformed, or contains empty content
- The required parameters aren't provided

To prevent the LLMChain from halting, the error gets returned but the malformed information isn't added to the LLMChain itself. This makes the LLMChain and tool calls more robust when using Mistral or other second/third tier LLMs.

## Changes

- `lib/function.ex`: Added parameter validation before function execution
- `lib/chains/llm_chain.ex`: Added logic to reject faulty message deltas
- `lib/message_delta.ex`: Refactored for better delta handling
- Added comprehensive tests for all new validation scenarios

## Test plan

- [x] All existing tests pass (1133 tests)
- [x] New tests added for parameter validation
- [x] New tests added for faulty message delta rejection

## ⚠️ Careful testing required

This has been tested with Mistral and it corrected Mistral successfully during API calls. However, I haven't tested this with other vendors. Be very careful upon merging this PR.

## Disclaimer

🤖 Co-Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/lib/message_delta.ex
+++ b/lib/message_delta.ex
@@ -241,7 +241,8 @@ defmodule LangChain.MessageDelta do
 
   # Empty content - nothing to merge
   defp append_to_merged_content(%MessageDelta{} = primary, %MessageDelta{content: content})
-       when content in [nil, []], do: primary
+       when content in [nil, []],
+       do: primary
 
   # String content - convert to ContentPart and merge
   defp append_to_merged_content(%MessageDelta{} = primary, %MessageDelta{content: content})


brainlid:
Thanks @xu-chris. This one makes me a little nervous. However, I made a point of releasing v0.5.0 yesterday so this could have more experimental time after merge. 

I appreciate the hardening. And I appreciate the work to help make the library more resilient to lower-end models and self-hosted models. Some of them are getting quite good for specific purposes.

I just want to call out that this will have a possible false positive with any model in a scenario like this:

**Scenario**: Assistant legitimately wants to respond with text starting with a tool name.

**Example**:
```
User: "What does the get_festival function do?"
Assistant: "get_festival returns information about..."
```

This would be flagged as malformed.

It's unlikely to pose a significant issue, but it's something to think about. It makes me wonder about pushing some of that responsibility back into the ChatModel.

Like a new ChatModel callback for validating or check_for_known_issues on an assistant response but letting the current model do that check. Just thinking about it.

xu-chris:
Thanks to you for establishing Langchain! It has helped me to build a robust and well working autonomous llm agent system. 

I totally understand your concerns and as I have stated, this has worked pretty well with the LLMs I was using but others may be less certain. I do not have enough money to spend to all LLMs for testing. Using streaming, parallel tool calls or even "run until tool used" does not make it easier.

Probably this is a community effort then. Let's use all our own preferences to use the library and see where the issues pop up.

brainlid:
@xu-chris I've been playing with this more recently (or rather, the updated code in LLMChain). I'm considering removing the changes to process_message. It creates a more complicated multi-path flow and ends up adding internally defined user messages that may show up for a user in a way that wouldn't be expected.

I'm releasing a new agent framework soon and the approach you're taking here would work better as a custom middleware that you add when you are talking to a chat model that may do this.

Just a heads up.

xu-chris:
@brainlid Thanks for playing a bit with this one and giving it a chance. It's absolutely fair to say that it's more part of the middleware than part of LangChain itself – would be probably more Elixir like and let the consumer decide how to handle weird messages by LLMs. 

I have thought it might be a good addition to LangChain to make it more a "batteries included" solution to what people are actually trying to accomplish – allow to offer the LLM some tools without having to deal with the LLM side effects. It already has the model normalization and an internal routing functionality with fallback models which I assumed is part of a middle layer. So my assumption was that LangChain could normalize the tool calls as well.

For me its fine if you exclude it again. I rewrote LangChain using ReqLLM instead which turned out to be even more stable. Only by your project I realized how I could do that as an Elixir newbie.

Thanks again and I am looking forward to your agent framework (since you've mentioned it the first time on Thinking Elixir podcast last year December)!

brainlid:
Thanks @xu-chris. I was thinking an alternative approach would be to make the message_processors work with tool calls. Originally, I introduced them for working with models that didn't support tool calls at all and needing to process a JSON result into a map and verify the structure and validity of it.

Glad you've found it helpful! Can you point me to your ReqLLM rewrite? I'm curious of the approach you took.

Issue #247, opened 2025-02-05
Retried streamed responses (ChatOpenAI) leak content from failed, retried requests


ci:
hey folks, noticed some weirdness where the response from OpenAI using streaming abruptly cuts off then gets regenerated (last tested on b349f18ded176c51add8630c8ead76436fdc6b26 but doesn't look like that changed in the meantime)

I've looked a bit through the logic, and I think the connection getting closed 'leaks' the previous messages by continuing the apply somehow, I repro'd (thanks to o3-mini based on a valid example from CF's AI gateway, hopefully it didn't botch the format) with the following:

<details>
<summary> client</summary>

```elixir
# client.exs
Mix.install([:req, :jason, :langchain])

alias LangChain.ChatModels.ChatOpenAI
alias LangChain.Message
alias LangChain.Chains.LLMChain

defmodule TestClient do
  def run do
    # Create a ChatOpenAI struct with our fake endpoint and streaming enabled.
    llm =
      ChatOpenAI.new!(%{
        endpoint: "http://localhost:31337/v1/chat/completions",
        stream: true,
        model: "gpt-4o",
        frequency_penalty: 0,
        temperature: 1,
        callbacks: []
      })

    # Build an LLMChain with a simple user message "test"
    chain =
      %{llm: llm}
      |> LLMChain.new!()
      |> LLMChain.add_messages([Message.new_user!("test")])

    updated_chain = LLMChain.run(chain, mode: :while_needs_response)

    {:ok, %LLMChain{last_message: %Message{content: content}}} = updated_chain

    IO.inspect(content, label: "CONTENT", printable_limit: :infinity)
  end
end

TestClient.run()
```

</details>

<details>
<summary>fake server closing connection</summary>

```elixir
# fake_server.exs
Mix.install([:plug_cowboy, :jason, :plug])

require Logger

defmodule FakeOpenAIServer do
  use Plug.Router

  plug(:match)
  plug(:dispatch)

  # Start an Agent to track call counts
  def start_agent do
    Agent.start_link(fn -> 0 end, name: __MODULE__.CallCounter)
  end

  post "/v1/chat/completions" do
    call_count =
      Agent.get_and_update(__MODULE__.CallCounter, fn count ->
        {count + 1, count + 1}
      end)

    Logger.info("Call count: #{call_count}")

    {:ok, body, conn} = Plug.Conn.read_body(conn)
    Logger.info("Request body: #{body}")
[...]

PR #450, opened 2026-01-27
fix(ChatOpenAIResponses): Handle failed response status from OpenAI Responses API

xu-chris:
## Problem

When the OpenAI Responses API returns a failed status, the current implementation does not handle it properly:

1. **Streaming mode**: Events with `"type": "response.failed"` fall through to the catch-all handler
2. **Non-streaming mode**: Responses with `"status": "failed"` also hit the catch-all

In both cases, users receive an unhelpful generic "Unexpected response" error instead of the actual error message from OpenAI (e.g., "Request timed out", "The server had an error processing your request").

Additionally, in streaming mode, the error was incorrectly wrapped in a list, causing a `CaseClauseError` downstream.

## Solution

Add explicit handlers for failed response status:

```elixir
# Streaming: handles "type": "response.failed" events
def do_process_response(_model, %{
      "type" => "response.failed",
      "response" => %{"status" => "failed"} = response
    })

# Non-streaming: handles responses with failed status
def do_process_response(_model, %{"response" => %{"status" => "failed"} = response})
```

Both handlers:
- Extract the actual error message from the response
- Return a properly structured `{:error, LangChainError}` tuple
- Log the error for debugging

## Based on PRs

- **PR #448**
- **PR #449**

## Related PRs/Issues

- **PR #392** - Improves the catch-all "Unexpected response" error by including `original`. This PR is complementary - it prevents failed responses from reaching the catch-all in the first place.
- **Issue #247** - Addresses content leaking from failed/retried requests. Different issue - that's about retry behavior, this is about proper error extraction.

## Changes

- `lib/chat_models/chat_open_ai_responses.ex` - Added two `do_process_response/2` clauses for failed status
- `test/chat_models/chat_open_ai_responses_test.exs` - Added tests for both streaming and non-streaming failed responses

## Test plan

- [x] All existing tests pass (1138 tests)
- [x] New tests for `response.failed` streaming event
- [x] New tests for failed response status (with and without error details)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,4 +32,5 @@ langchain-*.tar
 
 DEV_NOTES.md
 PERSONAL.md
-/logo/
+# macOS system files
+.DS_Store
--- a/lib/chat_models/chat_open_ai_responses.ex
+++ b/lib/chat_models/chat_open_ai_responses.ex
@@ -1284,32 +1284,12 @@ defmodule LangChain.ChatModels.ChatOpenAIResponses do
         "type" => "response.failed",
         "response" => %{"status" => "failed"} = response
       }) do
-    error_info = Map.get(response, "error", %{})
-    error_message = Map.get(error_info, "message", "Request failed")
-
-    Logger.error("OpenAI Responses API request failed: #{error_message}")
-
-    {:error,
-     LangChainError.exception(
-       type: "api_error",
-       message: "OpenAI request failed: #{error_message}",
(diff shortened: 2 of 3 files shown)


brainlid:
Thanks @xu-chris! The merge was hard to look at as it combined a previous PR and shows as many changes to Function that were more about shifting code locations. But I managed it.
❤️💛💙💜

PR #460, opened 2026-02-11
Fix exception handling in run_message_processors

Munksgaard:
The calling function, `process_message`, expects a 3-tuple in case of failures. This works for normal failures, but not for exceptions. This PR fixes that oversight, amends the existing test for `run_message_processors` and adds a new test for `process_message`.


brainlid:
Thanks @Munksgaard! I appreciate the fix.
❤️💛💙💜

PR #473, opened 2026-02-23
fix: handle nil content in run_message_processors

redmandarin:
Some OpenAI-compatible providers (e.g. Groq) omit the content field in assistant messages with tool calls, resulting in nil content. This caused a FunctionClauseError in parts_to_string/2 which requires a list argument.

Replaced parts_to_string with content_to_string which handles nil, string, and list content types. Defaults to empty string to ensure downstream processors can safely operate on processed_content.


brainlid:
Thanks @redmandarin!
❤️💛💙💜

PR #474, opened 2026-02-24
feat: add structured output support to ChatAnthropic

nallwhy:
## Summary
- Add `json_response` and `json_schema` fields to `ChatAnthropic` for Anthropic's native structured outputs API (`output_config.format`)
- Require `json_schema` when `json_response: true` via changeset validation, since Anthropic has no schema-less JSON mode unlike OpenAI
- No model allowlist — unsupported models will receive a clear API error, keeping the implementation future-proof

## References
- https://platform.claude.com/docs/en/build-with-claude/structured-outputs
- https://github.com/brainlid/langchain/pull/321


brainlid:
Nice @nallwhy! Thanks!
❤️💛💙💜

PR #478, opened 2026-02-27
feat(gemini): adding inline_data support for pdf and csv

conpat:
## Content of this PR

This PR adds support of inline files for PDFs and CSVs for Google Gemini (Relevant documentation is in the Code).
And tests for the above.

## My approach to the implementation

I tried to follow the same patterns as is used for:
- Anthropic
  https://github.com/brainlid/langchain/blob/cca047cb70b07c588d7a9764582773bae5703e8d/test/chat_models/chat_anthropic_test.exs#L2761:L2775
- Orq
  https://github.com/brainlid/langchain/blob/cca047cb70b07c588d7a9764582773bae5703e8d/lib/chat_models/chat_orq.ex#L400:L451

## Side notes

1. Although Google is vague about their support for anything else than PDFs ([docs](https://ai.google.dev/gemini-api/docs/file-input-methods)) we know it works because we already use this fork in production.
2. As of me writing this: I assume Google has a type in their [docs](https://ai.google.dev/gemini-api/docs/file-input-methods#text). I'm not aware of a mime type `text/scv`. I assume they mean `text/csv`.

### DIFF
--- a/lib/chat_models/chat_google_ai.ex
+++ b/lib/chat_models/chat_google_ai.ex
@@ -376,7 +376,8 @@ defmodule LangChain.ChatModels.ChatGoogleAI do
     }
   end
 
-  # Supported document types: pdf: https://ai.google.dev/gemini-api/docs/document-processing
+  # Supported document types: pdf: https://ai.google.dev/gemini-api/docs/document-processing#inline_data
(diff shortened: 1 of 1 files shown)


conpat:
Props to @xrty-oss 
I see that they did basically the same in #282

brainlid:
Thanks @conpat!
❤️💛💙💜

PR #483, opened 2026-03-10
feat: add WebSocket transport for ChatOpenAIResponses

nallwhy:
## Summary

- Add `LangChain.WebSocket`, a generic WebSocket client GenServer built on `mint_web_socket`, reusable across providers
- Add WebSocket transport option to `ChatOpenAIResponses` — when a WebSocket PID is set, requests go over the persistent connection instead of HTTP
- Add `connect_websocket!/1` and `disconnect_websocket/1` convenience API for pipe-friendly usage without manual URL/header setup
- Support both streaming and non-streaming modes over WebSocket

## Notes

- `temperature` and `top_p` are excluded from WebSocket payloads due to an [OpenAI bug](https://community.openai.com/t/responses-websocket-v1-responses-closes-with-code-1000-and-no-events-when-temperature-is-a-decimal-e-g-1-2/1375536) that silently closes the connection when these are sent as decimals. Can be restored once OpenAI fixes this.
- The `stream: false` + WebSocket combination may be unnecessary — WebSocket is inherently streaming, so removing the non-streaming WS path could be considered in a follow-up.

## Usage

```elixir
model =
  ChatOpenAIResponses.new!(%{model: "gpt-4o"})
  |> ChatOpenAIResponses.connect_websocket!()

{:ok, chain} =
  %{llm: model}
  |> LLMChain.new!()
  |> LLMChain.add_message(Message.new_user!("Hello"))
  |> LLMChain.run()

_model = ChatOpenAIResponses.disconnect_websocket!(model)
```

### DIFF
--- a/lib/chat_models/chat_open_ai_responses.ex
+++ b/lib/chat_models/chat_open_ai_responses.ex
@@ -898,8 +898,8 @@ defmodule LangChain.ChatModels.ChatOpenAIResponses do
                 events
                 |> Enum.find(&match?(%{"type" => "response.failed"}, &1))
                 |> case do
-                  %{"response" => response} ->
-                    do_process_response(openai, response)
+                  %{"type" => "response.failed"} = failed_event ->
+                    do_process_response(openai, failed_event)
 
                   nil ->
                     {:error,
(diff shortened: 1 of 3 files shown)


brainlid:
Thanks @nallwhy! It's merged in via #497.
❤️💛💙💜

I added additional documentation on websocket lifecycle behaviors and limitations.

PR #484, opened 2026-03-10
Handle empty LLM response in do_run/1 (thinking model streaming fix)

lurielle-studio:
## Summary

- Adds catch-all clauses to the `case message_response` block in `LLMChain.do_run/1` to handle `{:ok, []}` and other unexpected response formats
- Returns typed `LangChainError` instead of crashing with `CaseClauseError`

## Problem

When streaming with thinking/reasoning models (Qwen3, DeepSeek-R1, QwQ, etc.) via OpenAI-compatible APIs like OpenRouter, the model emits `<think>` tokens during its reasoning phase. The streaming parser in `handle_stream_fn` processes each chunk through `do_process_response`, which returns `:skip` for chunks with empty choices. After filtering skips, the response body accumulates as `[[], [], ...]`.

In `do_run/1`, the filter at line 840 rejects all empty lists:
```elixir
{:ok, Enum.reject(messages, &(&1 == []))}
```
This produces `{:ok, []}`, which doesn't match any clause in the subsequent `case`:
- `{:ok, [%Message{} = message]}` — needs at least one Message
- `{:ok, [%MessageDelta{} | _]}` — needs at least one MessageDelta
- etc.

Result: `CaseClauseError` crash.

## Fix

Two new clauses at the end of the `case`:

1. **`{:ok, []}`** — empty response, returns `LangChainError` with `type: "empty_response"`
2. **`{:ok, unexpected}`** — any other format, returns `LangChainError` with `type: "unexpected_response"`

## Test plan

- [x] Added test: empty response `{:ok, []}` returns structured error
- [x] Added test: unexpected format returns structured error
- [x] All 101 existing tests pass (6 live API tests excluded)

### DIFF
--- a/lib/chains/llm_chain.ex
+++ b/lib/chains/llm_chain.ex
@@ -888,6 +888,29 @@ defmodule LangChain.Chains.LLMChain do
         if chain.verbose, do: IO.inspect(string_reason, label: "ERROR")
         Logger.error("Error during chat call. Reason: #{inspect(string_reason)}")
         {:error, chain, LangChainError.exception(message: string_reason)}
+
+      {:ok, []} ->
+        # Empty response — all choices were filtered out (e.g., thinking model
+        # streaming where all chunks produce empty parsed results). Treat as
+        # an error rather than crashing with CaseClauseError.
+        Logger.warning("LLM returned an empty response (no messages or deltas)")
+
+        {:error, chain,
+         LangChainError.exception(
+           type: "empty_response",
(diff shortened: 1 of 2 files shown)


brainlid:
While this is still helpful, it's partly a symptom of valid content like `<think>` blocks getting filtered out when they shouldn't be. Perhaps that's another issue to tackle.

Thanks for the fix @lurielle-studio 
❤️💛💙💜

PR #489, opened 2026-03-25
Fix/vertex ai multimodal tool results

joechiarella:
Expands `ChatVertexAI` to support multimodal tool responses.

Vertex functionResponse has a very specific shape. 
- It has a single text or json "response", and zero-to-many media parts. 
  - Parts can not be text. 
- It's possible to reference a media part using a $ref block in the response json that matches the "displayName" property of the part. 
- Parts may be "inlineData" for base64 encoded media, or "fileData" for external references (generally a GCS uri). 

One example:

```json
"functionResponse": {
  "name": "name",
  "response": {
    "image": { "$ref": "image.png" }
  },
  "parts": [
    "inlineData": {
      "mimeType": "image/png",
      "data": "...",
      "displayName": "image.png"
    }
  ]
}
```

To map langchain content to this structure, all text parts are concatenated together to make the "response" and all media parts are added to the "parts" property. 

This was added in Gemini 3, but I don't think there are issues with backwards compatibility because attempting to return media parts would have always failed with earlier gemini versions.

Relevant docs: https://docs.cloud.google.com/vertex-ai/generative-ai/docs/multimodal/function-calling#mm-fr

### DIFF
--- a/lib/chat_models/chat_vertex_ai.ex
+++ b/lib/chat_models/chat_vertex_ai.ex
@@ -361,6 +361,16 @@ defmodule LangChain.ChatModels.ChatVertexAI do
 
   defp tool_result_response_for_api(nil), do: %{}
 
+  defp tool_result_response_for_api(content) when is_binary(content) do
+    case Jason.decode(content) do
+      {:ok, data} ->
+        data
+
+      {:error, %Jason.DecodeError{}} ->
+        %{"result" => content}
+    end
+  end
+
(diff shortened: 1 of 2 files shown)


brainlid:
Thank you for this work @joechiarella! We've incorporated your changes into #491, which also adds support for `:file` content parts (base64-encoded non-image files like PDFs) that were being silently dropped.

We're closing this PR in favor of #491 since maintainer push access wasn't enabled on this PR, which prevented us from pushing additional commits directly. The new PR preserves all of your original commits and credits your authorship.

PR #497, opened 2026-03-29
feat: add WebSocket transport for ChatOpenAIResponses

brainlid:
## Summary

Adds optional WebSocket transport to `ChatOpenAIResponses`, allowing requests to go over a persistent WebSocket connection instead of HTTP. Based on #483 by @nallwhy with bug fixes, documentation, and improvements.

- Introduces `LangChain.WebSocket`, a generic WebSocket client GenServer built on `mint_web_socket`, reusable across providers
- Adds `websocket` field to `ChatOpenAIResponses` with `connect_websocket!/1` and `disconnect_websocket!/1` convenience API
- Supports both streaming and non-streaming modes over WebSocket

### Changes from original PR (#483)

- **Bug fix**: `connect_websocket/1` had a no-op `unless` guard -- the error tuple for missing API keys was silently discarded, allowing the function to proceed with a nil key. Restructured as `if/else`.
- **Made `mint_web_socket` optional**: Since WebSocket is an opt-in feature, the dependency shouldn't be required for all consumers of the library.
- **Added Logger warning**: When `temperature` and `top_p` are dropped from WebSocket payloads (due to an [OpenAI bug](https://community.openai.com/t/1375536)), users now get a warning so their settings aren't silently ignored.
- **Cleaned up tests**: Replaced test that manually recreated private function logic with assertions through the actual code path via `do_api_request` mock.
- **Documented lifecycle management and limitations**: Both `LangChain.WebSocket` and `ChatOpenAIResponses` moduledocs now clearly explain that the application is responsible for managing the WebSocket lifecycle, and that the transport is not suited for long-lived agents, HITL workflows, or scenarios involving state serialization.

### Usage

```elixir
model =
  ChatOpenAIResponses.new!(%{model: "gpt-4o"})
  |> ChatOpenAIResponses.connect_websocket!()

{:ok, chain} =
  %{llm: model}
  |> LLMChain.new!()
  |> LLMChain.add_message(Message.new_user!("Hello"))
  |> LLMChain.run()

_model = ChatOpenAIResponses.disconnect_websocket!(model)
```

### Current Limitations

- **Lifecycle is application-managed**: The WebSocket GenServer is linked to the calling process with no supervisor, reconnection, or health monitoring. The PID stored in the model struct is not serializable -- if the struct is persisted to a database and restored, the WebSocket field will be stale. Best suited for short-lived, synchronous sessions.
- **No retry logic**: Unlike the HTTP transport, WebSocket failures are not retried. Mint/mint_web_socket intentionally leave retry logic to the application.
- **temperature/top_p dropped**: Due to an OpenAI bug that silently closes WebSocket connections when these are sent as decimals. A Logger warning is emitted.

### Notes

- The WebSocket transport works well for multi-turn conversations within a single uninterrupted session (e.g. `LLMChain.run(mode: :while_needs_response)` with tool calls).
- It is not safe for long-lived agent processes with human-in-the-loop interruptions or any scenario where the model struct crosses process boundaries or is serialized.

## Test plan

- [x] All 1597 tests pass, 0 failures
- [x] `mix precommit` clean
- [ ] Live WebSocket tests with OpenAI API (tagged `live_open_ai`, not run in CI)

Supersedes #483.

### DIFF
--- a/lib/chat_models/chat_open_ai_responses.ex
+++ b/lib/chat_models/chat_open_ai_responses.ex
@@ -194,6 +194,61 @@ defmodule LangChain.ChatModels.ChatOpenAIResponses do
 
       ChatOpenAIResponses.new!(%{model: "gpt-5", verbosity: "low"})
 
+  ## WebSocket Transport
+
+  Instead of HTTP, requests can be sent over a persistent WebSocket connection
+  for lower latency. Use `connect_websocket!/1` to open a connection and
+  `disconnect_websocket!/1` to close it:
+
+      model =
+        ChatOpenAIResponses.new!(%{model: "gpt-4o"})
+        |> ChatOpenAIResponses.connect_websocket!()
+
+      {:ok, chain} =
+        %{llm: model}
+        |> LLMChain.new!()
+        |> LLMChain.add_message(Message.new_user!("Hello"))
+        |> LLMChain.run()
+
+      ChatOpenAIResponses.disconnect_websocket!(model)
+
+  The WebSocket connection is reused across multiple LLM calls within the same
+  chain run (e.g. multi-turn tool calling with `:while_needs_response`).
+
+  ### Lifecycle Management
+
+  **The application is responsible for managing the WebSocket lifecycle.**
+  `connect_websocket!/1` starts a `LangChain.WebSocket` GenServer via
+  `start_link/1`, linking it to the calling process. The PID is stored in
+  the model struct's `:websocket` field. There is no supervisor, automatic
+  reconnection, or health monitoring built in.
+
+  This means:
+
+  - If the calling process exits, the WebSocket is terminated (process link).
+  - The WebSocket PID cannot be serialized. If the model struct is persisted
+    to a database and restored later, the `:websocket` field will be stale.
+  - The server may close idle connections at any time. There is no automatic
+    reconnection.
+  - There is no retry logic for WebSocket failures (unlike the HTTP transport).
+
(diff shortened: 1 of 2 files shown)


brainlid:
## WebSocket Lifecycle Note

One thing to be aware of with this WebSocket transport: the `LangChain.WebSocket` GenServer is started via `start_link/1` and linked to the calling process. The PID is stored directly in the `ChatOpenAIResponses` struct's `:websocket` field.

This design works well for short-lived, synchronous sessions where the caller controls the full lifecycle. However, it has implications for more complex usage patterns:

- **Process linking**: If the process that called `connect_websocket!` exits, the WebSocket is terminated. In architectures where LLM calls happen in spawned tasks (e.g. `Task.async`), the WebSocket would die when the task completes.
- **Not serializable**: The PID can't survive serialization. If the model struct is saved to a database and restored later (common in agent frameworks with state persistence), the `:websocket` field will be a stale PID.
- **No reconnection**: If OpenAI closes the idle connection server-side, subsequent calls return `{:error, :not_connected}` with no automatic recovery.
- **No retry logic**: Unlike the HTTP path which has retry support, WebSocket failures fail immediately.

I've added documentation to both `LangChain.WebSocket` and `ChatOpenAIResponses` making these limitations explicit and recommending the default HTTP transport for long-lived or interruptible workloads.

Future improvements could include adding the WebSocket to a supervision tree, implementing reconnection logic, or having `do_api_request` fall back to HTTP when the WebSocket is dead. But for a first version, clearly documenting the current behavior seems like the right approach.

PR #502, opened 2026-04-01
Handle {:interrupt, chain, data} and {:pause, chain} in try_chain_with_llm

### LINKED ISSUE #501: try_chain_with_llm doesn't handle {:interrupt, chain, data} — a declared Mode.run_result() type
## Bug

`LLMChain.try_chain_with_llm/4` only matches `{:ok, _}` and `{:error, _, _}` in its `case` statement ([llm_chain.ex L706-722](https://github.com/brainlid/langchain/blob/main/lib/chains/llm_chain.ex#L706-L722)), but the `Mode` behaviour declares `{:interrupt, LLMChain.t(), term()}` as a valid `run_result()` type ([mode.ex L103](https://github.com/brainlid/langchain/blob/main/lib/chains/llm_chain/mode.ex#L103)).

When a custom mode (e.g. Sagents' `AgentExecution`) returns `{:interrupt, chain, data}` and fallback models are configured, the unmatched tuple causes a `CaseClauseError`. The `rescue` block catches it and retries on the next fallback — which also interrupts — until all fallbacks are exhausted, returning a misleading `"all_fallbacks_failed"` error.

This makes **fallback models and HITL interrupts mutually exclusive**, which is problematic in production: transient LLM errors (Bedrock auth failures, rate limits) can't be recovered via fallbacks when HITL is active.

## Error

```
** (CaseClauseError) no case clause matching: {:interrupt, %LangChain.Chains.LLMChain{...}}
```

Logged as:
```
Rescued from exception during with_fallback processing. Error: %CaseClauseError{term: {:interrupt, %LangChain.Chains.LLMChain{...}}}
```

## Suggested fix

Add an `{:interrupt, _, _}` clause to `try_chain_with_llm` that passes through immediately (it's not an error to retry):

```elixir
defp try_chain_with_llm(%LLMChain{} = chain, [llm | tail], before_fallback_fn, run_fn) do
  # ... existing setup ...
  try do
    case run_fn.(use_chain) do
      {:ok, result} ->
        {:ok, result}

      {:interrupt, _chain, _data} = interrupt ->
        interrupt

      {:error, _error_chain, reason} = error ->
        if llm_module.retry_on_fallback?(reason) do
          try_chain_with_llm(use_chain, tail, before_fallback_fn, run_fn)
        else
          error
        end
    end
  rescue
    # ... existing rescue ...
  end
end
```

The same treatment should be applied to `{:pause, chain}` since it is also a declared `run_result()` type.

## Environment

- langchain 0.6.3
- sagents 0.4.3
- Elixir 1.15+

srikanthkyatham:
## Summary

- Add explicit `case` clauses for `{:interrupt, _, _}` and `{:pause, _}` in `try_chain_with_llm/4` so they pass through immediately instead of causing a `CaseClauseError`
- Previously, these valid `Mode.run_result()` types were caught by the `rescue` block and incorrectly retried on fallback LLMs, making fallback models and HITL interrupts mutually exclusive
- Add test coverage for both new clauses with custom test mode modules

Closes #501

## Test plan

- [x] Added test: `with_fallbacks: passes through {:interrupt, chain, data} without retrying`
- [x] Added test: `with_fallbacks: passes through {:pause, chain} without retrying`
- [x] All 1611 existing tests continue to pass (`mix precommit`)

### DIFF
--- a/lib/chains/llm_chain.ex
+++ b/lib/chains/llm_chain.ex
@@ -705,8 +705,11 @@ defmodule LangChain.Chains.LLMChain do
 
     try do
       case run_fn.(use_chain) do
-        {:ok, result} ->
-          {:ok, result}
+        {:ok, _chain} = ok ->
+          ok
+
+        {:ok, _chain, _extra} = ok ->
+          ok
 
         {:interrupt, _chain, _data} = interrupt ->
           # Interrupt is a deliberate pause for human-in-the-loop or similar
--- a/test/chains/llm_chain_test.exs
+++ b/test/chains/llm_chain_test.exs
@@ -2396,6 +2396,26 @@ defmodule LangChain.Chains.LLMChainTest do
       assert %LLMChain{} = returned_chain
     end
 
+    test "with_fallbacks: passes through {:ok, chain, extra} without retrying" do
+      # The custom mode returns {:ok, chain, extra} (3-tuple) without calling the LLM.
+      # This is used by Steps.check_until_tool when a target tool is found.
+
+      chain =
+        %{llm: ChatOpenAI.new!(%{stream: false})}
+        |> LLMChain.new!()
+        |> LLMChain.add_message(Message.new_system!())
+        |> LLMChain.add_message(Message.new_user!("Do something"))
+
+      {:ok, returned_chain, returned_extra} =
+        LLMChain.run(chain,
+          mode: LangChain.Test.OkExtraMode,
+          with_fallbacks: [ChatAnthropic.new!(%{stream: false})]
+        )
+
+      assert %LLMChain{} = returned_chain
+      assert returned_extra == %{tool_result: "found_it"}
+    end
+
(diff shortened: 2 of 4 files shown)


brainlid (approved):
Good catch! The fix for `{:interrupt, _, _}` and `{:pause, _}` is correct.

I pushed one additional commit that:

1. **Also handles `{:ok, chain, extra}` (3-tuple)** -- this is another declared `run_result()` type that wasn't matched. The `Steps.check_until_tool/2` function returns `{:ok, chain, tool_result}` when a target tool is found, and the original `{:ok, result}` clause only matches the 2-tuple variant. Same class of bug, same fix.

2. **Split test mode modules into separate files** -- one module per file per project convention.

All 1612 tests pass.

brainlid:
Thanks @srikanthkyatham! Great catch!
❤️💛💙💜

PR #507, opened 2026-04-02
fix: generate unique call_id per tool call in ChatGoogleAI

raghusubramanyam:
## Summary
- `ChatGoogleAI` was constructing `call_id` as `"call-#{tool_name}"` — not unique per invocation
- When Gemini issues parallel calls to the same tool, all invocations shared the same `call_id`
- Caused `Ecto.MultipleResultsError` in AEGIS when `Repo.one` found 2 pending display messages with the same `call_id`
- Fix: use `Ecto.UUID.generate()` so each tool call invocation gets a unique ID (consistent with Perplexity and Ollama adapters)

## Test plan
- [ ] New test: "same tool called twice produces distinct call_ids" — asserts `tc1.call_id != tc2.call_id`
- [ ] All 82 ChatGoogleAI tests pass

### DIFF
--- a/lib/chains/llm_chain.ex
+++ b/lib/chains/llm_chain.ex
@@ -880,7 +880,12 @@ defmodule LangChain.Chains.LLMChain do
         {:ok, updated_chain}
 
       {:ok, []} ->
-        Logger.error("LLM returned empty response — no messages or deltas")
+        Logger.error(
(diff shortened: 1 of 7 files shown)


raghusubramanyam:
Closing — this fix is internal to our AEGIS fork, not intended for upstream.

PR #532, opened 2026-05-01
adding support for thinking to ollama models

nolman:
debugging some models locally and would like to see the reasoning the models reasoning - letting the thinking pass through langchain


brainlid:
Nice @nolman! Thanks!
❤️💛💙💜

PR #540, opened 2026-05-07
ChatOpenAIResponses: Fix streaming parser for SSE event boundary splitting

montebrown:
Streaming responses from OpenAI's Responses API started failing — typically only the first delta arrived before all subsequent events stopped parsing, no further callbacks fired, and chains hung indefinitely. Non-streaming mode was unaffected.

Root cause: decode_stream/2 used regex splitting on "event: " which assumed event and data lines always arrived in the same TCP chunk. A change in OpenAI's TCP chunking behavior broke this assumption, causing the event name to become an invalid JSON fragment in the buffer that cascaded to corrupt all subsequent event parsing.

Rewrote to split on "\n\n" (the actual SSE event boundary), then extract the "data:" line from each complete event block. This handles all chunking patterns: split events, multiple events per chunk, and incomplete JSON across chunks.


brainlid:
Thanks @montebrown! That's a really important one. Personally, I don't use that endpoint much myself, so I rely on people like you to help maintain it. I appreciate your help!
❤️💛💙💜

PR #541, opened 2026-05-07
ChatAnthropic: Add output_config field for thinking effort control

montebrown:
Anthropic's API accepts output_config as a top-level request parameter that controls output behavior such as adaptive thinking effort.

- Added output_config :map field to the embedded schema
- Added :output_config to @create_fields so it passes through cast
- Added set_output_config/1 clause to include it in the API request body


brainlid:
Thanks @montebrown!
❤️💛💙💜

PR #542, opened 2026-05-07
LLMChain: Improve error handling for unexpected response formats in do_run/1

montebrown:
Added catch-all clauses in do_run/1 to handle edge cases that previously caused crashes or confusing errors:

- {:ok, [[error: %LangChainError{}] | _]} — API errors wrapped in an :ok tuple (e.g., Anthropic overloaded_error during streaming)
- {:error, reason} — non-string, non-LangChainError error reasons

Each case now returns a structured {:error, chain, reason} tuple with an appropriate error type and message.


brainlid:
Thanks @montebrown!
❤️💛💙💜

PR #543, opened 2026-05-07
ToolCall: Fix crash when merging binary arguments into already-decoded map

montebrown:
During streaming, merge/2 calls append_arguments before update_status. When a delta arrives with status: :complete and a decoded map for arguments, append_arguments replaces the primary's binary arguments with the map. update_status then marks the primary as :complete. If the provider sends a trailing delta with a binary JSON fragment for the same tool call, append_arguments tried to concatenate the binary onto the map via <>, which crashes with an ArgumentError.

We observed this in production — some providers send overlapping or trailing deltas after the tool call is already complete. Since the arguments are already fully decoded, there is no meaningful way to merge a partial JSON string into the map. The fragment is safely discarded.

Added a guard clause that detects map arguments with an incoming binary and returns the primary unchanged, preventing the crash.

### DIFF
--- a/lib/message/tool_call.ex
+++ b/lib/message/tool_call.ex
@@ -239,11 +239,25 @@ defmodule LangChain.Message.ToolCall do
     %ToolCall{primary | arguments: new_arguments}
   end
 
-  defp append_arguments(%ToolCall{} = primary, %ToolCall{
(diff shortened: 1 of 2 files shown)


brainlid:
Thanks @montebrown! Very nice!
❤️💛💙💜

PR #556, opened 2026-05-18
Add :origin option to FileGoogle.request_upload_url

nallwhy:
## Problem

`FileGoogle.request_upload_url/4` initiates Google's two-step resumable upload and returns a presigned URL. It's exposed as a module function so callers can forward the URL to a browser/mobile client and let it perform step 2 directly.

When step 2 runs in a browser, CORS applies. Google sets `Access-Control-Allow-Origin` on the step 2 response based on the `Origin` header sent on step 1 — but there's no first-class way to set that from `request_upload_url/4`. Callers have to inject it through the `req_opts` escape hatch, which relies on hidden knowledge about Google's upload protocol that domain-meaningful options like `:display_name` already shield them from.

(This is Google-specific. OpenAI ([docs](https://developers.openai.com/api/reference/resources/files/methods/create)) and Anthropic ([docs](https://platform.claude.com/docs/en/build-with-claude/files)) Files APIs use a single multipart POST with no presigned-URL flow, so there's no shared shape to lift into the `FileUploader` behaviour.)

## Solution

Extend `request_upload_url/4` to `request_upload_url/5` with a trailing `opts \\ []`. When `opts[:origin]` is set, inject it as the `"origin"` header on the step 1 request.

The default `[]` keeps the existing `upload/3` call site source-compatible. Keyword opts (rather than a positional `origin`) leave room for future options. `upload_file_bytes/4` is deliberately not extended — step 2 either runs server-side (no CORS) or in a browser (which sets `Origin` automatically).

### DIFF
--- a/lib/file_uploaders/file_google.ex
+++ b/lib/file_uploaders/file_google.ex
@@ -25,6 +25,20 @@ defmodule LangChain.FileUploader.FileGoogle do
   available for use with `ContentPart.file_url!/2`.
 
   Note: Files uploaded to Gemini expire after 48 hours.
+
+  ## Client-side direct upload
+
+  Step 2 can be performed by a browser or mobile client instead of your server.
+  Call `request_upload_url/5` with the client's origin and forward the returned
+  URL to the client — the client then PUTs the file bytes to that URL directly.
+
+      {:ok, upload_url} =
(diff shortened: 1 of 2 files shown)


brainlid:
Thanks @nallwhy!
❤️💛💙💜

PR #569, opened 2026-06-22
Ignore SSE comment lines in ChatOpenAI.decode_stream (fixes #259)

### LINKED ISSUE #259: Ignore SSE comments
For streaming, OpenAI-compatible APIs send back a SSE stream. This stream can include comments (line starting with a colon) which can be ignored per the [SSE Specs](https://html.spec.whatwg.org/multipage/server-sent-events.html#event-stream-interpretation). [OpenRouter](https://openrouter.ai/docs/api-reference/streaming#additional-information) especially seems to be sending such comments.
> For SSE (Server-Sent Events) streams, OpenRouter occasionally sends comments to prevent connection timeouts.

Below is an example that failed for me.

## Notes

For reference, this is how the [Python OpenAI SDK](https://github.com/openai/openai-python/blob/2e56c8da6f163db00a4ca362020148bb391edca9/src/openai/_streaming.py#L343) handles it.

Not sure whether this is the proper way, but this solved it for me: https://github.com/n18abdel/langchain/commit/a5db5cee49a7fc71c8b9dd41689a9356f9cad1c3

## Example

```elixir
alias LangChain.ChatModels.ChatOpenAI
alias LangChain.Message
alias LangChain.MessageDelta

Application.put_env(:langchain, :openai_key, "<API_KEY>")

model = "meta-llama/llama-3-8b-instruct:free"
endpoint = "https://openrouter.ai/api/v1/chat/completions"

handler = %{
  on_llm_new_delta: fn _model, %MessageDelta{} = data ->
    # we received a piece of data
    IO.write(data.content)
  end,
  on_message_processed: fn _chain, %Message{} = data ->
    # the message was assembled and is processed
    IO.puts("")
    IO.puts("")
    IO.inspect(data.content, label: "COMPLETED MESSAGE")
  end
}

{:ok, updated_chain} =
  %{llm: ChatOpenAI.new!(%{model: model, endpoint: endpoint, stream: true})}
  |> LLMChain.new!()
  |> LLMChain.add_messages([
    Message.new_system!("You are a helpful assistant."),
    Message.new_user!("Write a haiku about the capital of the United States")
  ])
  |> LLMChain.add_callback(handler)
  |> LLMChain.run()

updated_chain.last_message.content
```

## Error

```elixir
** (CaseClauseError) no case clause matching: {:ok, []}
    (langchain 0.3.1) lib/chains/llm_chain.ex:552: LangChain.Chains.LLMChain.do_run/1
    (langchain 0.3.1) lib/chains/llm_chain.ex:398: LangChain.Chains.LLMChain.run/2
```

a-designer:
Fixes #259.

## Problem

OpenAI-compatible streaming APIs may send SSE **comment** lines (starting with a colon), which are ignorable per the [SSE spec](https://html.spec.whatwg.org/multipage/server-sent-events.html#event-stream-interpretation). OpenRouter in particular sends `: OPENROUTER PROCESSING` keep-alive comments to prevent connection timeouts.

`ChatOpenAI.decode_stream/2` splits only on `"data: "`, so a comment line is treated as JSON, fails to decode, and is buffered as `incomplete`. That buffer is then prepended to every subsequent valid chunk — poisoning it — so the entire stream parses to `[]`. Downstream this surfaces as an empty response (and, before the `{:ok, []}` handling was added, a `CaseClauseError` in `LLMChain.do_run/1`).

## Fix

Skip lines beginning with `":"` in the `decode_stream` reduce, mirroring how the [Python OpenAI SDK](https://github.com/openai/openai-python/blob/2e56c8da6f163db00a4ca362020148bb391edca9/src/openai/_streaming.py#L343) handles SSE comments.

## Tests

Adds two regression tests to `decode_stream/2`:
- a lone `": OPENROUTER PROCESSING"` comment parses to `[]` with no leftover buffer
- a data chunk that follows a comment still parses correctly (proves the buffer isn't poisoned)

Full `ChatOpenAITest` suite passes (109 tests). Verified live against OpenRouter: streaming now assembles content token-by-token instead of returning empty.

### DIFF
--- a/lib/chat_models/chat_open_ai.ex
+++ b/lib/chat_models/chat_open_ai.ex
@@ -1055,6 +1055,13 @@ defmodule LangChain.ChatModels.ChatOpenAI do
       str
       |> String.trim()
       |> case do
+        ":" <> _sse_comment ->
+          # A line starting with a colon is an SSE comment and can be ignored per
+          # https://html.spec.whatwg.org/multipage/server-sent-events.html#event-stream-interpretation
+          # OpenRouter sends ": OPENROUTER PROCESSING" keep-alive comments which
+          # otherwise poison the incomplete-JSON buffer and break the whole stream.
+          acc
+
         "" ->
           acc
 
--- a/test/chat_models/chat_open_ai_test.exs
+++ b/test/chat_models/chat_open_ai_test.exs
@@ -2080,6 +2080,29 @@ defmodule LangChain.ChatModels.ChatOpenAITest do
       assert parsed == [json_1, json_2]
     end
 
+    test "ignores SSE comment lines (e.g. OpenRouter keep-alives)" do
+      # OpenRouter sends SSE comments (lines starting with ":") to keep the
+      # connection alive. They are ignorable per the SSE spec and must not be
+      # buffered as incomplete JSON, which would poison all following chunks.
+      data = ": OPENROUTER PROCESSING\n\n"
+
+      {parsed, incomplete} = ChatOpenAI.decode_stream({data, ""})
+
+      assert incomplete == ""
+      assert parsed == []
+    end
+
+    test "parses a data chunk that follows an SSE comment", %{json_1: json_1} do
+      data =
+        ": OPENROUTER PROCESSING\n\n" <>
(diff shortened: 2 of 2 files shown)


brainlid:
Thanks @a-designer!
❤️💛💙💜

brainlid:
Fix released in v0.8.14

PR #575, opened 2026-07-02
fix: do not put the Vertex AI API key in the request URL

pinetops:
## Problem

`ChatVertexAI.build_url/1` appends the API key as a query parameter:

```elixir
"#{endpoint}/models/#{model}:#{get_action(vertex_ai)}?key=#{get_api_key(vertex_ai)}"
```

The same key is *also* already sent as a bearer token (`auth: {:bearer, get_api_key(vertex_ai)}` in the Req options), so the query-string copy is redundant - it looks like it arrived via copy paste from ChatGoogleAI. It is also a credential-leak vector: full request URLs routinely appear in server/proxy access logs, error trackers, and browser history — exposing the API key.

## Fix

Drop `?key=...` from the URL. Authentication is unchanged — the bearer token in the `Authorization` header still carries the key.

Also matched test approach based on ChatGoogleAI

### DIFF
--- a/lib/chat_models/chat_vertex_ai.ex
+++ b/lib/chat_models/chat_vertex_ai.ex
@@ -622,14 +622,15 @@ defmodule LangChain.ChatModels.ChatVertexAI do
     end
   end
 
+  @doc false
   @spec build_url(t()) :: String.t()
-  defp build_url(%ChatVertexAI{endpoint: endpoint, model: model} = vertex_ai) do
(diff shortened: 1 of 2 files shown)


brainlid:
@pinetops is this ready for review? I usually don't look at drafts.

brainlid:
Thanks @pinetops!
❤️💛💙💜

PR #576, opened 2026-07-02
feat: support configurable safety_settings on ChatVertexAI

pinetops:
## Summary

`ChatGoogleAI` exposes a `safety_settings` field passed through to the API as `safetySettings`, but `ChatVertexAI` has no equivalent — callers can't set Vertex AI harm-category thresholds, so the provider defaults always apply.

This adds a `safety_settings` field to `ChatVertexAI` (default `[]`), mirroring `ChatGoogleAI`:
- cast it in `@create_fields`
- conditionally add it to the request body as `safetySettings`

## Behaviour

Default `[]` is a no-op (an empty list is not added to the request), so existing behaviour is unchanged. Callers who need to tune harm thresholds — e.g. a stricter or more permissive filter for their domain — can now do so:

```elixir
ChatVertexAI.new!(%{
  model: "...",
  endpoint: "...",
  safety_settings: [
    %{"category" => "HARM_CATEGORY_DANGEROUS_CONTENT", "threshold" => "BLOCK_ONLY_HIGH"}
  ]
})
```

## Test

Mirrors the existing `ChatGoogleAI` safety-settings tests: present settings are sent as `safetySettings`; an empty list is omitted. `chat_vertex_ai_test.exs` passes (41 tests).

### DIFF
--- a/lib/chat_models/chat_vertex_ai.ex
+++ b/lib/chat_models/chat_vertex_ai.ex
@@ -116,6 +116,14 @@ defmodule LangChain.ChatModels.ChatVertexAI do
     # Refer to `https://hexdocs.pm/req/Req.html#new/1-options` for
     # `Req.new` supported set of options.
     field :req_config, :map, default: %{}
+
+    # The safety settings for the model, specified as a list of maps. Each map
+    # should contain a `category` and a `threshold` for that category.
+    # e.g. [%{"category" => "HARM_CATEGORY_DANGEROUS_CONTENT", "threshold" => "BLOCK_ONLY_HIGH"}]
+    # see https://cloud.google.com/vertex-ai/generative-ai/docs/multimodal/configure-safety-filters
+    # for the list of categories and thresholds. Defaults to `[]` (the
+    # provider's own defaults apply).
+    field :safety_settings, {:array, :map}, default: []
   end
 
   @type t :: %ChatVertexAI{}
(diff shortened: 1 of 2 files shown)


brainlid:
Hi @pinetops! I see the PR is still a draft. Please let me know when it's ready. Thanks!

brainlid:
Thanks for this @pinetops, and sorry for the slow turnaround.

The change is good and I've carried it forward in #593 with your commit intact, so you keep authorship of the feature. I opened a separate PR rather than pushing here because the head repo is org-owned (`u2i/langchain`) and "allow edits by maintainers" doesn't grant push access on org-owned forks, so I couldn't update your branch directly.

One thing I added on top: `serialize_config/1` builds its output from an explicit allowlist passed to `Utils.to_serializable_map/3`, and `:safety_settings` wasn't in it, so a serialize/restore round-trip dropped the configured thresholds and the restored model quietly fell back to the provider defaults. `ChatGoogleAI` already lists it in its own `serialize_config`, so adding it is what gets the two to real parity. Also rebased onto `main` and added a round-trip test plus a CHANGELOG entry.

Easy thing to miss, and not a knock on the PR. Closing in favor of #593.

brainlid:
Thanks @pinetops! The change was merged in via https://github.com/brainlid/langchain/pull/593
❤️💛💙💜

PR #431, opened 2026-01-05
feat(GoogleChatAI): add thought_signature support for Gemini 3 function calls

abh1shek-sh:
The Gemini 3 family of models requires [thought_signatures](https://ai.google.dev/gemini-api/docs/thought-signatures) in order to do a tool call. This information provides Google with the reasoning why the tool call was made and apparently is mandatory in order for tool calls to work. 

This PR adds a thought_signature support for Gemini 3 models, it's not required for older models though.

PR #587, opened 2026-07-23
Fix ChatReqLLM dropping Anthropic thinking signatures on tool-loop replay

mstroeck:
We hit this in production: with extended thinking enabled, every chat turn where Claude calls a tool fails. Turn 1 streams back a thinking block plus a tool_use. When the chain sends the follow-up request with the tool result, the assistant message gets replayed with its thinking block and Anthropic rejects it:

```
messages.1.content.0.thinking.signature: Field required
status: 400, type: invalid_request_error
```

Anthropic requires replayed thinking blocks to carry the signature from the original response. ChatReqLLM loses that signature at three points:

1. Streaming: req_llm accumulates `signature_delta` internally and only exposes it in a terminal meta chunk carrying `reasoning_details` (see agentjido/req_llm#343 and agentjido/req_llm#626). `process_stream_chunk/2` ignored that chunk, so the merged assistant message never had the signature.
2. Non-streaming: req_llm puts the signature on `message.reasoning_details`, not on the thinking content part, so `req_llm_content_part_to_lc/1` found no `metadata[:signature]` either.
3. Outbound: `content_part_to_req_llm/1` serialized thinking parts as bare text and dropped `options[:signature]` even when present.

Worth knowing why nobody noticed earlier: req_llm versions before 1.17.1 silently dropped thinking parts from outgoing Anthropic requests, which the API tolerates. 1.17.1 started encoding them properly, so the unsigned blocks began reaching Anthropic and thinking + tools stopped working entirely.

The fix follows what ChatAnthropic already does (#523 established the omit-unsigned behavior, and #431 did the same signature round-trip for Gemini): keep the signature in `options[:signature]` on the thinking ContentPart, pass it back as ReqLLM part metadata on serialization, and omit unsigned thinking blocks with a warning instead of sending one the API will refuse.

Covered by new unit tests, and verified against the live Anthropic API with thinking + tool call + continuation in both stream modes. We have been running this patch in production since 2026-07-21 with thinking re-enabled.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixed
+
+- **`ChatReqLLM` now round-trips Anthropic extended-thinking signatures.** With thinking enabled, Anthropic requires assistant thinking blocks to be replayed with their original signature on tool-result continuations. The adapter lost the signature on the way in and on the way out: streaming never picked it up from the terminal `reasoning_details` meta chunk, non-streaming responses carry it on `message.reasoning_details` rather than the content part, and `content_part_to_req_llm/1` dropped `options[:signature]` on serialization. The result was a 400 (`thinking.signature: Field required`) on every thinking + tool-use turn. Signatures now survive both response paths and are passed back as part metadata; unsigned thinking blocks are omitted with a warning, same as `ChatAnthropic`.
+
 ### Added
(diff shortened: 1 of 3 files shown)


brainlid:
Thanks @mstroeck! Great catch and fix.
❤️💛💙💜

PR #589, opened 2026-07-24
Expose the executed LLMChain from DataExtractionChain

brainlid:
## Problem

`DataExtractionChain.run/4` returns only the extracted data. There is no way to reach the underlying `LLMChain`, so callers can't log or report the token usage of an extraction.

This was raised in #588 by @Chuukwudi, who needed token usage for NER tasks.

## Why not `return_message: true`

#588 solves it with a `return_message: true` option on `run/4`. That works and is non-breaking, but it makes the shape of the `{:ok, _}` return depend on a boolean flag, which flattens the typespec to effectively `{:ok, term()}` for every call site — including the ones that never pass the flag. It also only returns the message on the happy path where the extraction tool call matched, so usage is unreachable in exactly the failure case you'd most want to inspect.

## Approach

The other helper chains in this repo already have a shape for this: `run/N` returns the executed `LLMChain`, and a second function reduces it to the domain value.

| | returns the chain | returns the value |
|---|---|---|
| `RoutingChain` | `run/2` | `evaluate/2` |
| `TextToTitleChain` | `run/2` | `evaluate/2` |
| `SummarizeConversationChain` | `run/3` | `summarize/3`\* |
| `DataExtractionChain` (before) | — | `run/4` |

\* `summarize/3` folds the summary back into a conversation chain rather than returning a bare value, but `run/3` has the same return shape as the others.

This PR gives `DataExtractionChain` the missing half:

- **`run_chain/4`** returns `{:ok, %LLMChain{}} | {:error, %LLMChain{}, %LangChainError{}}`, matching `LLMChain.run/2`.
- **`extract_result/1`** pulls the extracted data out of an executed chain.
- **`run/4`** is unchanged in signature and return, and is now implemented as `run_chain/4 |> extract_result/1`, so the default path and the escape hatch cannot drift apart.

```elixir
{:ok, chain} = DataExtractionChain.run_chain(chat, schema_parameters, data_prompt)

usage = LangChain.TokenUsage.get(chain.last_message)
{:ok, result} = DataExtractionChain.extract_result(chain)
```

Because `run_chain/4` returns the chain whether or not the model produced the expected extraction tool call, token usage stays reportable even when the extraction itself fails.

The `try/rescue` stays on `run/4` only, wrapping the call to `run_chain/4`, so `run/4`'s existing error behavior is preserved. `run_chain/4` lets `PromptTemplate.to_messages!` and `LLMChain.new!` raise, consistent with `RoutingChain.run/2` and `TextToTitleChain.run/2`.

## Tests

Added unit tests (Mimic-mocked, no live calls) covering `extract_result/1` on both the matching and non-matching chain, `run_chain/4` returning usage in the success and the no-tool-call case, and `run/4`'s unchanged behavior.

Full suite: 35 doctests, 2035 tests, 0 failures. `mix format` clean, compiles with `--warnings-as-errors`.

> Note: `mix precommit` currently fails at `mix deps.audit` on `main` as well, due to advisories against `mint 1.7.1` (`< 1.9.0`). Unrelated to this change — flagging it separately.

## Alternative to #588

If this lands, #588 can be closed. Credit to @Chuukwudi for identifying the gap.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 ### Added
 
+- **`DataExtractionChain` exposes the executed `LLMChain` through `run_chain/4` and `extract_result/1`.** `run/4` returns only the extracted data, so there was no way to reach the underlying chain to log or report the token usage of an extraction. `run_chain/4` returns `{:ok, %LLMChain{}}` — matching how `RoutingChain`, `TextToTitleChain`, and `SummarizeConversationChain` already work — and `extract_result/1` pulls the extracted data out of it. Because the chain is returned regardless of whether the model produced the expected tool call, usage is reportable even when the extraction itself fails. `run/4` is unchanged and is now implemented in terms of the two.
 - **`ChatAnthropic` accepts `suppress_api_key: true` to omit the `x-api-key` header.** When the Anthropic Messages endpoint is fronted by its own authentication — a corporate proxy, or an AWS SigV4-signed gateway such as Bedrock "Mantle" — sending `x-api-key` is rejected or ignored. Setting `suppress_api_key: true` suppresses the header entirely so the caller's own auth (e.g. `req_opts` SigV4 signing) can take over. Defaults to `false`, preserving existing behavior.
 
 ## v0.9.2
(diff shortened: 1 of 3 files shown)


Chuukwudi:
@brainlid Perfect. Thank you very much. I like this approach better !!!

brainlid:
> @brainlid Perfect. Thank you very much. I like this approach better !!!

Thanks @Chuukwudi! This revealed another opportunity to improve observability of these operations. That will come as a follow-up PR.

PR #591, opened 2026-07-24
build(deps): update mint and req to clear security advisories

brainlid:
## Problem

`mix precommit` currently fails on `main` at the `deps.audit` step. `mix_audit` reports four advisories against `mint` 1.7.1, so the project's own pre-commit gate has been red for everyone.

## Change

Lockfile only. No `mix.exs` constraint changed, since `{:req, ">= 0.5.3"}` already permits 0.6.

| package | from | to |
|---|---|---|
| `mint` | 1.7.1 | 1.9.3 |
| `req` | 0.5.17 | 0.6.3 |
| `finch` | 0.21.0 | 0.23.0 |
| `hpax` | 1.0.3 | 1.0.4 |
| `decimal` | 3.1.0 | 3.1.1 |

### What the advisories were

**`mint`**, the four that `deps.audit` was flagging:

- [GHSA-g586-ccqf-7x4r](https://github.com/advisories/GHSA-g586-ccqf-7x4r) (high): unbounded `streams` map growth via `PUSH_PROMISE` frames with no follow-up `HEADERS`.
- [GHSA-2pg6-44cx-c49v](https://github.com/advisories/GHSA-2pg6-44cx-c49v): CRLF injection into the HTTP request line via unvalidated `method` / `target`.
- [GHSA-mjqx-c6f6-7rc2](https://github.com/advisories/GHSA-mjqx-c6f6-7rc2): `Content-Length` parser accepting a non-RFC `+` sign prefix.

**`req`** came along for `finch ~> 0.21`, but it turns out to matter on its own. req 0.6.0 includes [GHSA-px9f-whj3-246m](https://github.com/wojtekmach/req/security/advisories/GHSA-px9f-whj3-246m), which escapes `name`, `filename`, and `content_type` in multipart part headers per RFC 7578. This library hits that path directly:

```elixir
# lib/file_uploaders/file_open_ai.ex
filename = Map.fetch!(file_meta, :filename)
...
|> Req.post(
  form_multipart: [
    {"purpose", purpose},
    {"file", {file_bytes, filename: filename, content_type: mime_type}}
  ]
)
```

`:filename` is caller-supplied and was interpolated into the part headers unescaped, so a value containing `"` or CRLF could inject extra headers or smuggle additional parts. Same shape in `FileAnthropic.upload/2`. So this bump closes a real vector in LangChain's own upload path, not just a transitive advisory.

## req 0.6 behavior changes, and why they are safe here

Two are worth knowing about. I checked both against the codebase and neither applies:

1. **Response decompression is now opt-in** (`compressed: true`, req 0.6.1). LangChain never sets `:compressed` and never relies on `decompress_body`. Req simply stops advertising `accept-encoding: gzip`, so conforming servers return identity-encoded bodies that parse as before.
2. **`decode_body` decodes only JSON by default** (req 0.6.0); zip/tar/gz/zst/csv moved behind the new `:decoders` option. LangChain only ever decodes JSON, which is still on by default.

`grep` for `compressed`, `decompress`, `decode_body`, and `decoders` across `lib/` returns nothing but an unrelated comment.

## Verification

`mix precommit` passes end to end, which it does not on `main`:

```
No vulnerabilities found.
...
35 doctests, 2044 tests, 0 failures (143 excluded)
```

Compiles with `--warnings-as-errors`, `sobelow` and `mix format` clean. The 58 tests in `test/file_uploaders/` covering the multipart paths pass unchanged.

One caveat worth stating plainly: the unit suite mocks at the chat-model layer, so it exercises little real `req` code. The 143 excluded `live_call` tests are the ones that would put req 0.6 against actual provider endpoints. Worth a live run before release.


brainlid:
## Dialyzer failure diagnosed and fixed in ba00995

The lint job failed at `mix dialyzer`. Root cause is an upstream type change in mint, not anything in this repo's code.

**mint 1.9.0 demoted `Mint.HTTP.t()` from an opaque type of its own to a transparent alias:**

```elixir
# mint 1.7.1
@opaque t() :: Mint.HTTP1.t() | Mint.HTTP2.t()

# mint 1.9.3
@type t() :: Mint.HTTP1.t() | Mint.HTTP2.t() | Mint.UnsafeProxy.t()
```

While it was opaque, passing a conn from `Mint.HTTP.connect/4` into `Mint.WebSocket.upgrade/4` moved one sealed type between modules and dialyzer was satisfied. Now the alias expands, so dialyzer sees the concrete opaque terms of `Mint.HTTP1` / `Mint.HTTP2` / `Mint.UnsafeProxy` crossing `LangChain.WebSocket`'s boundary:

```
lib/web_socket.ex:283:48:call_with_opaque
The call Mint.WebSocket.upgrade(...) contains an opaque term in 2nd argument
when terms of different types are expected in these positions
```

`Mint.WebSocket.upgrade/4` still declares its `conn` as `Mint.HTTP.t()`, so this trips for any caller doing the documented `connect` then `upgrade` sequence.

**There is no code-side fix.** `do_connect/1` already passes `protocols: [:http1]`, so narrowing with `{:ok, %Mint.HTTP1{} = conn}` looks tempting, but matching that struct from outside its module is itself an opacity violation. It would trade one warning for another.

So this adds a `:call_with_opaque` filter next to the `lib/web_socket.ex` entry already in `.dialyzer_ignore.exs`, with the type change documented inline. That is the same treatment the file already gives the ReqLLM spec issue.

### Verified locally

| | errors | skipped | result |
|---|---|---|---|
| `main` (CI) | 14 | 14 | pass |
| this branch before the fix | 11 | 10 | **fail**, 1 emitted |
| this branch after the fix | 11 | 11 | **pass** |

The error count dropping 14 to 11 is also mint's doing: several previously-suppressed `pattern_match` warnings in `web_socket.ex` no longer fire under the new types.

> Unrelated, and pre-existing on `main`: both runs report `Unnecessary Skips: 1`, so one filter in `.dialyzer_ignore.exs` no longer matches anything. Left alone as out of scope here.

PR #592, opened 2026-07-24
fix: accept callbacks through chat model new/1

brainlid:
## Problem

`:callbacks` is declared as a schema field on every chat model, but was missing from `@create_fields` on 12 of the 14. `Ecto.Changeset.cast/3` drops any param that isn't in the permitted list, so this:

```elixir
{:ok, chat} = ChatOpenAI.new(%{callbacks: [handlers]})
```

returns `{:ok, model}` with `callbacks: []`, no error, no warning. Any handler registered that way never fires.

This is worse than an inconsistency, because both `ChatOpenAI` and `ChatOpenAIResponses` document exactly that call in their **Rate Limit API Response Headers** section:

```elixir
handlers = %{
  on_llm_ratelimit_info: fn _model, headers ->
    IO.inspect(headers)
  end
}

{:ok, chat} = ChatOpenAI.new(%{callbacks: [handlers]})
```

So the documented way to read OpenAI rate-limit headers could not work. The other ten modules never documented it, but were silently inconsistent with `ChatGrok` and `ChatOrq`, which already permit the field, as does `LLMChain`. That existing precedent is what makes this a bug rather than a design choice.

Found while working on #590, where handlers set on an `llm` turned out to be unreachable for a second, unrelated reason.

## Change

Added `:callbacks` to `@create_fields` in the 12 affected modules:

`ChatAnthropic`, `ChatAwsMantle`, `ChatBumblebee`, `ChatDeepSeek`, `ChatGoogleAI`, `ChatMistralAI`, `ChatOllamaAI`, `ChatOpenAI`, `ChatOpenAIResponses`, `ChatPerplexity`, `ChatReqLLM`, `ChatVertexAI`

Each diff is one line. Setting callbacks by struct update, which is what every test in the repo does, was never affected and keeps working:

```elixir
chat = %ChatOpenAI{chat | callbacks: [handler]}
```

Also corrected the example in `chat_open_ai_responses.ex`, which called `ChatOpenAI.new/1` rather than `ChatOpenAIResponses.new/1`.

## Tests

The regression test in `chat_model_test.exs` is table-driven over every chat model that declares a `:callbacks` field, rather than one test per module. A new provider that forgets to permit the field fails the test instead of shipping the same silent drop.

It has two parts, and the second is what keeps the first honest:

1. every listed model accepts `:callbacks` through `new/1`
2. every listed model constructs from its minimum attrs and still defaults `:callbacks` to `[]`

Without (2), a model whose `new/1` merely returned `{:error, changeset}` for unrelated reasons would be indistinguishable from one that accepted callbacks correctly. While building the fixture, that check caught three models needing real attrs (`ChatAwsMantle`, `ChatMistralAI`, `ChatVertexAI`) that would otherwise have been counted as failures for the wrong reason.

Verified the test fails before the fix, listing exactly the 12 affected modules and correctly excluding `ChatGrok` and `ChatOrq`.

## Verification

- `mix precommit` passes end to end: 35 doctests, 2046 tests, 0 failures.
- `mix dialyzer` run locally: `Total errors: 11, Skipped: 11`, `done (passed successfully)`, unchanged from `main`.


brainlid:
## Follow-up in f32ed4d: the fix above was only half of it

Verifying that the newly-accepted callbacks actually *fire* turned up two more problems. Both are addressed here, along with an integration test that pins the behavior down.

### 1. `LLMChain` was discarding model callbacks

`Utils.rewrap_callbacks_for_model/3` forwards the chain's LLM-level handlers down onto the model for the duration of a run, but it replaced the model's list rather than adding to it:

```elixir
# put those onto the model and return it
%{llm | callbacks: new_callbacks}
```

So `ChatOpenAI.new(%{callbacks: [handler]})` was accepted by the first commit and then silently dropped again the moment the model ran inside a chain, which is the path most callers use. Now appends, so both sets run.

The chain's forwarded maps are built only from `chain.callbacks`, never from the model's, so appending cannot double-fire a model handler. There's a test for that specifically rather than leaving it to reasoning.

### 2. The documented handler arity fails the whole call

The two tiers take different arguments. A chat model fires its own handlers with just the event's argument; `LLMChain` curries itself in as a first argument. That split is visible in the existing tests:

```elixir
# model tier, chat_open_ai_test.exs:1179
on_llm_new_message: fn %Message{} = new_message -> ... end        # arity 1

# chain tier, chat_open_ai_test.exs:1297
on_llm_new_delta: fn %LLMChain{} = _chain, deltas -> ... end      # arity 2
```

But `ChatOpenAI`, `ChatOpenAIResponses`, and `ChatGrok` all documented a **two**-argument handler assigned to the *model*:

```elixir
handlers = %{on_llm_ratelimit_info: fn _model, headers -> IO.inspect(headers) end}
{:ok, chat} = ChatOpenAI.new(%{callbacks: [handlers]})
```

Following that produces a `BadArityError`, which `Callbacks.fire/3` re-raises as a `LangChainError`, so the entire LLM call fails:

```
{:error, %LangChainError{message: "Callback handler for :on_llm_ratelimit_info raised an
  exception: (BadArityError) #Function<...>/2 with arity 2 called with 1 argument (%{})"}}
```

Worth flagging: **this is already reachable on `main` for `ChatGrok`**, which has always permitted `:callbacks`. For the OpenAI modules it was masked only because the field was being dropped, so the first commit here would have turned a silent no-op into a broken client. Corrected to one argument in all three, with a note explaining both tiers. `ChatAnthropic`'s example was already correct, since it registers on the chain via `add_callback`.

`LLMChain`'s moduledoc also claimed its handlers are "not provided to an LLM chat model", which was wrong in both directions. It now documents both tiers.

### Integration test

`test/callbacks_integration_test.exs` drives the real `ChatOpenAI.call/3` against a mocked `Req.post/1`, so the provider's own callback firing is exercised rather than stubbed out:

| case | asserted |
|---|---|
| model callbacks, direct `call/3` | fire |
| model callbacks, through `LLMChain` | fire (the bug) |
| chain callbacks, through `LLMChain` | fire |
| both assigned | both fire |
| model handler | not called twice |
| chain's list handed to the model | model's own first, chain's appended, both arity 1 |
| arity-2 handler on a model | fails the call, pinning the trap |

I wrote these against the *old* behavior first: they asserted the wipe, passed, and then the merge change flipped exactly those three and no others. The full suite was run against the merge prototype before committing it and **0 of the 2046 existing tests changed behavior**, which is the main evidence that appending is safe.

### Verification

`mix precommit` clean: 35 doctests, **2054 tests, 0 failures**. Leaving dialyzer to CI on this one.

PR #277, opened 2025-03-23
Support for json_response in ChatModels.ChatGoogleAI

nallwhy:
This PR introduces support for the json_response option in `ChatModels.ChatGoogleAI`. With this update, users can now enable JSON response formatting by setting `json_response` to true. When enabled, the API request is enhanced with a `response_mime_type` of `"application/json"` and, if a `json_schema` is provided, it is included in the request payload.

Issue #210, opened 2024-12-09
Supporting structured JSON output for Google AI


vkryukov:
Looks like Google supports structured JSON output by supplying a schema [through model configuration](https://ai.google.dev/gemini-api/docs/structured-output?lang=rest):
```
curl "https://generativelanguage.googleapis.com/v1beta/models/gemini-1.5-flash:generateContent?key=$GOOGLE_API_KEY" \
-H 'Content-Type: application/json' \
-d '{
    "contents": [{
      "parts":[
        {"text": "List 5 popular cookie recipes"}
        ]
    }],
    "generationConfig": {
        "response_mime_type": "application/json",
        "response_schema": {
          "type": "ARRAY",
          "items": {
            "type": "OBJECT",
            "properties": {
              "recipe_name": {"type":"STRING"},
            }
          }
        }
    }
}' 2> /dev/null | head
```

Would you be interested in a PR that adds this ability, similar to ChatOpenAI [implementation](https://github.com/brainlid/langchain/blob/315e787c7f4e52c014b52bda0142aa631d3dd28f/lib/chat_models/chat_open_ai.ex#L290) (which btw should be documented?)

brainlid:
Hi @vkryukov! Yes, I'd love help implementing and supporting Google Gemini support for structured output with a schema declaration. :smile:

vkryukov:
Sounds good, I will try develop a PR to start as a discussion point.

Google's models are improving fast (both gemini-2.0-flash-exp and gemini-exp-1206 are pretty impressive), and they offer massive context windows - so I want to keep testing them. So we need to have a first-class support by `LangChain`.

brainlid:
That's cool. I've been so underwhelmed and disappointed by Google's past models that I wrote them off for personal use. Glad to hear they are improving!

vkryukov:
Yeah, I can relate to that. I plan to post a comparison (eventually); for my use case (code generation), `gemini-exp-1206` is not bad! But of course, Sonnet 3.5 New still seems to be the SOTA.

jjmilburn:
Seconding that this functionality would be very useful.

It looks like Ash AI (which uses this project) supports a `StructuredOutput` adapter (https://hexdocs.pm/ash_ai/AshAi.Actions.Prompt.Adapter.StructuredOutput.html) which only supports OpenAI for now, but first-class support for Gemini structured output (https://ai.google.dev/gemini-api/docs/structured-output) in Langchain would likely change that.

PR #595, opened 2026-07-25
docs: document structured JSON output for ChatGoogleAI and ChatVertexAI

pinetops:
Addresses the documentation gap from #210.

## Background

#210 asked for structured JSON output support in `ChatGoogleAI`, and noted the equivalent OpenAI feature "should be documented". The **feature itself already shipped** — `json_response` + `json_schema` (mapping to the Gemini API's `generationConfig.response_mime_type` / `response_schema`) are implemented and tested in both `ChatGoogleAI` (#277) and `ChatVertexAI`. But it was **undocumented**, so it was undiscoverable and its usage unclear.

## Change

Adds a `## Structured Output` moduledoc section to `ChatGoogleAI` and `ChatVertexAI`, each with:

- a worked `json_schema` example (the cookie-recipes schema from the issue),
- a link to [Google's structured-output docs](https://ai.google.dev/gemini-api/docs/structured-output),
- the mapping to `response_mime_type` / `response_schema`,
- a note that `json_response: true` works with or without a `json_schema`, and that the assistant content is a JSON string to decode.

Docs-only — no behavioural change. Uses a current model (`gemini-2.5-flash`) in the examples.

Once this lands, #210 can be closed (feature implemented in #277, now documented).


brainlid:
Hi @pinetops! Note with the recent merging of https://github.com/brainlid/langchain/pull/599, schema changes Google APIs have changed. There's a deny list and suppression of the keys that caused problems before. Just mentioning in case it's more directly related to this.

brainlid:
The suggested changes and an additional change could not be applied to this PR branch because it belongs to an organization. 

The final set of changes were competed and merged in https://github.com/brainlid/langchain/pull/610.

PR #597, opened 2026-07-27
Recover string tool-result content in ChatGoogleAI serialization

### LINKED ISSUE #596: ChatGoogleAI raises FunctionClauseError on a ToolResult whose content is a string
## Summary

`ToolResult.content` accepts a string or a list of ContentParts. Every other
provider that serializes a tool result handles both shapes. `ChatGoogleAI` does
not: its `ToolResult` clause calls `ContentPart.parts_to_string/1`, which is
guarded `when is_list(parts)`, so a string raises `FunctionClauseError` while
the request is being built.

The failure is persistent rather than transient. The crash happens while
serializing the conversation history, so once such a result is in the history,
every later turn re-serializes it and fails the same way. The conversation
cannot continue.

Verified on `main` (cbd3acd) and on v0.9.3.

## Use case, and why this matters

Any code that sets a tool result's content directly hits this. The shape is
easy to reach without knowing it: `ToolResult.new/1` runs
`Utils.migrate_to_content_parts/1`, which wraps a string into a list, so a
result built through the constructor is always a list and always works. A
struct built directly, or an existing result updated with `%ToolResult{r |
content: "..."}`, keeps the string. Both are supported shapes, and the second
is the ordinary way to rewrite a result that already exists — for example when
a held or deferred tool call is answered later and its result is filled in.

What makes this worth fixing in the library rather than working around is that
the same code works on ChatAnthropic and fails on ChatGoogleAI. A caller who
develops against one provider and switches to the other gets a crash from
library-internal serialization, with nothing in the typespecs or the docs to
say the shape was disallowed.

## Context

The shape is legal by four separate pieces of the library's own contract:

1. `ToolResult`'s moduledoc states it directly: "The `content` can be a string
   or a list of ContentParts for multi-modal responses (text, images, etc.)
   that gets returned to the LLM as the result." The field is
   `field :content, :any, virtual: true`.
2. `ContentPart.content_to_string/2` is the accessor written for exactly this
   union. Its spec is `content :: String.t() | [t()] | nil`, and it has an
   explicit `is_binary` clause.
3. Other providers already serialize it. `ChatDeepSeek` and `ChatOllamaAI` use
   `content_to_string` on this path; `ChatAnthropic.content_parts_for_api/1`
   and `ChatOrq.content_parts_to_string/1` each carry an explicit `is_binary`
   clause.
4. `ChatGoogleAI` already handles binary content elsewhere in the same module —
   for `:system` message content, and for `Message` content — through
   `when is_binary(content)` clauses. Only the `ToolResult` clause was missed.

Point 4 is what makes this look like an oversight rather than a decision. The
module is internally inconsistent with itself: two of its three content paths
accept a string, and the third crashes on it.

`ChatMistralAI`, `ChatOllamaAI`, and `ChatPerplexity` have `parts_to_string`
calls on paths that can receive a binary too. I have not audited whether each
is reachable, so I am reporting them rather than claiming them.

## Reproduction

1. Build a `ToolResult` whose `content` is a plain string, without going
   through `ToolResult.new/1`.
2. Serialize it with `ChatGoogleAI.for_api/1` — or, equivalently, put it in a
   conversation and run any turn.

```elixir
alias LangChain.ChatModels.ChatGoogleAI
alias LangChain.Message.ToolResult

result = %ToolResult{
  name: "find_theaters",
  tool_call_id: "call-find_theaters",
  content: "I don't know where the theaters are."
}

ChatGoogleAI.for_api(result)
```

```
** (FunctionClauseError) no function clause matching in
   LangChain.Message.ContentPart.parts_to_string/2

    The following arguments were given:
        # 1
        "I don't know where the theaters are."
        # 2
        :text

    Attempted function clauses (showing 1 out of 1):
        def parts_to_string(parts, type) when is_list(parts)

    (langchain) lib/message/content_part.ex:327
    (langchain) lib/chat_models/chat_google_ai.ex:417
```

The same result serializes without error on `ChatAnthropic.for_api/2`.

## Observed

`ChatGoogleAI.for_api/1` raises `FunctionClauseError` on a documented and
supported shape of `ToolResult.content`. Because the raise happens during
history serialization, the conversation is unusable from that point on — every
subsequent turn fails identically.

## Expected

`ChatGoogleAI` serializes a string `content` the same way the other providers
do, producing the same request it would produce for the equivalent single-text
ContentPart list.

## Offer

I have a fix prepared and can open a PR. It replaces the `parts_to_string/1`
call with `ContentPart.content_to_string/1`, which already accepts nil, a
string, or a list — the same helper `ChatDeepSeek` and `ChatOllamaAI` use on
this path. It is a one-line change plus a regression test that builds the
struct directly, since going through `new!/1` is what hides the bug.

I am happy to take a different shape if you prefer one — an explicit
`is_binary` clause matching the `ChatAnthropic` style would work equally well.
I can also extend the same fix to `ChatMistralAI`, `ChatOllamaAI`, and
`ChatPerplexity` in the same PR or a separate one, whichever you would rather
review.

lostbean:
Fixes #596.

`ToolResult.content` may be a string or a list of ContentParts. The moduledoc
states both shapes, `ContentPart.content_to_string/2` is typed for both, and
`ChatGoogleAI` itself already handles both for `:system` and `Message` content
through `when is_binary(content)` clauses.

Its `ToolResult` clause did not. It called `ContentPart.parts_to_string/1`,
which is guarded `when is_list(parts)`, so a string raised
`FunctionClauseError` while the request was being serialized.

`ToolResult.new/1` runs `Utils.migrate_to_content_parts/1`, which wraps a
string into a list, so a result built through the constructor never hit this. A
struct built directly, or an existing result updated with
`%ToolResult{r | content: "..."}`, keeps the string, and that is a supported
shape. The failure is persistent rather than transient: the raise happens while
serializing the conversation history, so once such a result is in the history,
every later turn fails the same way and the conversation cannot continue.

The fix uses `content_to_string/1`, which accepts nil, a string, or a list.
That is the same helper `ChatDeepSeek` and `ChatOllamaAI` already use on this
path, and it matches what `ChatAnthropic.content_parts_for_api/1` and
`ChatOrq.content_parts_to_string/1` do through an explicit `is_binary` clause.
I chose it over adding a separate `is_binary` clause because it is the accessor
written for this union, and it keeps the clause a single expression. Happy to
switch to an explicit clause if you prefer the `ChatAnthropic` style.

The regression test builds the `ToolResult` struct directly rather than through
`new!/1`, because the constructor's migration is exactly what hides the bug — a
test written the usual way passes against the unpatched code. It asserts the
string case produces the same map the equivalent ContentPart list produces. I
confirmed it fails with `FunctionClauseError` without the fix.

`ChatMistralAI`, `ChatOllamaAI`, and `ChatPerplexity` have `parts_to_string`
calls on paths that can receive a binary as well. I left them alone to keep
this PR to one provider, and can follow up on them separately if you want them
covered.

### DIFF
--- a/lib/chat_models/chat_google_ai.ex
+++ b/lib/chat_models/chat_google_ai.ex
@@ -412,9 +412,11 @@ defmodule LangChain.ChatModels.ChatGoogleAI do
   end
 
   def for_api(%ToolResult{} = result) do
+    # `content` may be a string or a list of ContentParts. `content_to_string/1`
+    # accepts both, where `parts_to_string/1` requires a list.
     content_string =
       result.content
-      |> ContentPart.parts_to_string()
+      |> ContentPart.content_to_string()
 
     content =
       content_string
--- a/lib/chat_models/chat_google_ai.ex
+++ b/lib/chat_models/chat_google_ai.ex
@@ -412,24 +412,13 @@ defmodule LangChain.ChatModels.ChatGoogleAI do
   end
 
   def for_api(%ToolResult{} = result) do
-    # `content` may be a string or a list of ContentParts. `content_to_string/1`
-    # accepts both, where `parts_to_string/1` requires a list.
-    content_string =
+    # `content` may be nil, a string, or a list of ContentParts.
+    # `content_to_string/1` accepts all three, where `parts_to_string/1`
+    # requires a list.
+    content =
       result.content
       |> ContentPart.content_to_string()
-
-    content =
-      content_string
-      |> Jason.decode()
-      |> case do
-        {:ok, data} ->
-          # content was converted through JSON
-          data
-
-        {:error, %Jason.DecodeError{}} ->
-          # assume the result is intended to be a string and return it as-is
-          %{"result" => content_string}
-      end
+      |> tool_result_response_for_api()
 
     # There is no explanation for why they want it nested like this. Odd.
     #
@@ -468,6 +457,24 @@ defmodule LangChain.ChatModels.ChatGoogleAI do
 
   def for_api(%NativeTool{name: name, configuration: nil}) do
     name
+  end
+
+  # A tool result may carry no text at all: `content` can be nil, or a
+  # multi-modal ContentPart list holding only non-text parts. Send an empty
+  # object for those, matching ChatVertexAI, rather than handing nil to
+  # `Jason.decode/1`.
+  defp tool_result_response_for_api(nil), do: %{}
+
+  defp tool_result_response_for_api(content_string) when is_binary(content_string) do
+    case Jason.decode(content_string) do
+      {:ok, data} ->
+        # content was converted through JSON
+        data
+
+      {:error, %Jason.DecodeError{}} ->
+        # assume the result is intended to be a string and return it as-is
+        %{"result" => content_string}
+    end
   end
 
   @doc """
--- a/test/chat_models/chat_google_ai_test.exs
+++ b/test/chat_models/chat_google_ai_test.exs
@@ -358,6 +358,50 @@ defmodule ChatModels.ChatGoogleAITest do
       assert expected == ChatGoogleAI.for_api(tool_result)
     end
 
+    test "tool result with no text content creates expected map" do
+      # A multi-modal tool result may carry no `:text` part at all. Reached
+      # through the normal `new!/1` constructor, so this is not limited to
+      # hand-built structs.
+      expected = %{
+        "functionResponse" => %{
+          "name" => "take_screenshot",
+          "response" => %{
+            "name" => "take_screenshot",
+            "content" => %{}
+          }
+        }
+      }
+
+      tool_result =
+        ToolResult.new!(%{
+          name: "take_screenshot",
+          tool_call_id: "call-take_screenshot",
+          content: [ContentPart.image!("base64data", media: :png)]
+        })
+
+      assert expected == ChatGoogleAI.for_api(tool_result)
+    end
+
+    test "tool result with nil content creates expected map" do
+      expected = %{
+        "functionResponse" => %{
+          "name" => "find_theaters",
+          "response" => %{
+            "name" => "find_theaters",
+            "content" => %{}
+          }
+        }
+      }
+
+      tool_result = %ToolResult{
+        name: "find_theaters",
+        tool_call_id: "call-find_theaters",
+        content: nil
+      }
+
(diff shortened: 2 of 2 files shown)


brainlid:
Thanks, this is a good catch and the fix is the right shape. `content_to_string/1` is the accessor written for this union, and `ChatVertexAI` reaches the same Gemini API and already handles both shapes through explicit `tool_result_response_for_api/1` clauses. So `ChatGoogleAI` was the outlier here rather than the two providers disagreeing about intent.

I confirmed the regression: your test raises `FunctionClauseError` against the unpatched clause and passes with the fix.

I pushed a commit on top instead of asking for changes, because the same function has a second crash in the same class. Once `content_to_string/1` accepts the union, a `content` that yields no text still reaches `Jason.decode(nil)`:

```
** (ArgumentError) errors were found at the given arguments:

  * 1st argument: not an iodata term

    (erts 17.0.1) :erlang.iolist_to_binary(nil)
    (jason 1.4.5) lib/jason.ex:69: Jason.decode/2
    (langchain) LangChain.ChatModels.ChatGoogleAI.for_api/1
```

Two ways to land there: `content: nil`, and a multi-modal ContentPart list holding no `:text` part. The second is the interesting one, because it goes through the normal constructor rather than a directly built struct:

```elixir
ToolResult.new!(%{
  tool_call_id: "call-1",
  name: "take_screenshot",
  content: [ContentPart.image!("base64data", media: :png)]
})
```

That is exactly what the `ToolResult` moduledoc advertises the list shape for, so any tool returning an image to Gemini hits it. The commit pulls the decode into a `tool_result_response_for_api/1` helper with a `nil -> %{}` clause, mirroring `ChatVertexAI`'s helper of the same name, and adds tests for both cases. `mix precommit` is clean.

On the other providers you flagged, I checked all of the call sites:

* `ChatMistralAI` and `ChatPerplexity` are already safe. Every `parts_to_string` call there is either in an `is_list`-guarded clause head or inside a `case` with an `is_binary` branch, and `content_for_tool_result/2` has explicit binary, list, and nil clauses.
* `ChatOllamaAI` does have the identical defect, in the same place:

  ```elixir
  def for_api(%ToolResult{content: content}) do
    %{"role" => :tool, "content" => ContentPart.parts_to_string(content)}
  end
  ```

  Unguarded, so a string raises the same way.

Agreed on keeping Ollama out of this PR. A separate one for it would be welcome if you want to take it.

PR #599, opened 2026-07-27
Strip JSON-Schema keywords Gemini's function declarations reject

### LINKED ISSUE #440: Tool schemas missing `additionalProperties: false` causing errors with Anthropic API
## Summary

When using tools/function calling with `ChatAnthropic`, requests fail with error:
```
tools.0.custom: For 'object' type, 'additionalProperties' must be explicitly set to false
```

This is because `LangChain.FunctionParam.to_parameters_schema/1` generates JSON schemas for tools without the required `additionalProperties: false` field that Anthropic's API strictly requires.

## Environment

- **LangChain version**: 0.4.1
- **Anthropic model**: `claude-sonnet-4-5-20250929` (also affects all Claude models)
- **Elixir version**: 1.17+
- **Integration**: Used via AshAi framework, but affects any direct usage

## Steps to Reproduce

1. Define a tool with parameters using LangChain:

```elixir
alias LangChain.ChatModels.ChatAnthropic
alias LangChain.Function

# Create a simple tool
tool = Function.new!(%{
  name: "get_weather",
  description: "Get the weather for a location",
  parameters: [
    FunctionParam.new!(%{
      name: "location",
      type: :string,
      required: true,
      description: "The city name"
    }),
    FunctionParam.new!(%{
      name: "unit",
      type: :string,
      enum: ["celsius", "fahrenheit"],
      description: "Temperature unit"
    })
  ]
})

# Use with ChatAnthropic
{:ok, chain} =
  %{llm: ChatAnthropic.new!(%{model: "claude-3-5-sonnet-20241022"})}
  |> LLMChain.new!()
  |> LLMChain.add_tools([tool])
  |> LLMChain.add_message(Message.new_user!("What's the weather in Paris?"))
  |> LLMChain.run()
```

2. Anthropic API returns error:

```
[error] Received error from API: %{
  "error" => %{
    "message" => "tools.0.custom: For 'object' type, 'additionalProperties' must be explicitly set to false",
    "type" => "invalid_request_error"
  }
}
```

## Root Cause

The issue is in `lib/function_param.ex` line 222-228:

```elixir
def to_parameters_schema(params) when is_list(params) do
  %{
    "type" => "object",
    "properties" => Enum.reduce(params, %{}, &to_json_schema(&2, &1)),
    "required" => required_properties(params)
    # MISSING: "additionalProperties" => false
  }
end
```

Additionally, nested object parameters (line 269-276) also don't include this field:

```elixir
def to_json_schema(%{} = data, %FunctionParam{type: :object, object_properties: props} = param) do
  settings =
    props
    |> to_parameters_schema()  # This recursive call also misses additionalProperties
    |> description_for_schema(param.description)

  Map.put(data, param.name, settings)
end
```

## Expected Behavior

Generated schemas should include `"additionalProperties": false` for all object types to comply with Anthropic's API requirements:

```json
{
  "type": "object",
  "properties": {
    "location": {
      "type": "string",
      "description": "The city name"
    },
    "unit": {
      "type": "string",
      "enum": ["celsius", "fahrenheit"],
      "description": "Temperature unit"
    }
  },
  "required": ["location"],
  "additionalProperties": false
}
```

## Actual Behavior

Current schema generation omits `additionalProperties`:

```json
{
  "type": "object",
  "properties": { ... },
  "required": ["location"]
}
```

## Anthropic API Requirements

From [Anthropic's API documentation](https://docs.anthropic.com/en/docs/build-with-claude/tool-use):

> For object type parameters, `additionalProperties` must be explicitly set to `false`

This is a strict requirement enforced by their API validation.

### Fix 1: Update `to_parameters_schema/1`

```elixir
def to_parameters_schema(params) when is_list(params) do
  %{
    "type" => "object",
    "properties" => Enum.reduce(params, %{}, &to_json_schema(&2, &1)),
    "required" => required_properties(params),
    "additionalProperties" => false  # ADD THIS LINE
  }
end
```

### Fix 2: Handle existing schemas with `additionalProperties`

In `chat_anthropic.ex` around line 1759-1762, when a schema is provided directly:

```elixir
defp get_parameters(%Function{parameters: [], parameters_schema: schema} = _fun)
     when is_map(schema) do
  # Ensure additionalProperties is set for Anthropic compatibility
  Map.put_new(schema, "additionalProperties", false)
end
```

### Fix 3: Add to empty object schemas

In `chat_anthropic.ex` line 1752-1757:

```elixir
defp get_parameters(%Function{parameters: [], parameters_schema: nil} = _fun) do
  %{
    "type" => "object",
    "properties" => %{},
    "additionalProperties" => false  # ADD THIS LINE
  }
end
```

## Impact

- **OpenAI**: Not affected (their API is more permissive)
- **Anthropic**: All tool/function calling is broken
- **Other providers**: May vary depending on their schema validation strictness

## Workaround

Manually edit `deps/langchain/lib/function_param.ex` to add the field (not recommended for production).

Alternatively, use `ChatOpenAI` when tools are needed, as OpenAI doesn't enforce this requirement.

## Additional Context

This issue was discovered while integrating LangChain with the AshAi framework for tool calling. The error manifests when `AshAi.setup_ash_ai/2` adds tools to the chain, which internally uses `LangChain.FunctionParam.to_parameters_schema/1` to generate schemas.

The fix should be backward compatible as adding `additionalProperties: false` is valid JSON Schema and aligns with Anthropic's strict requirements without affecting other providers.

## References

- [Anthropic Tool Use Documentation](https://docs.anthropic.com/en/docs/build-with-claude/tool-use)
- [JSON Schema Specification](https://json-schema.org/understanding-json-schema/reference/object#additionalproperties)
- [Anthropic API Models](https://docs.anthropic.com/en/docs/about-claude/models)

### LINKED ISSUE #598: ChatGoogleAI sends JSON Schema keywords Gemini rejects, so every tool with parameters fails
## Summary

Google's function declarations accept a select subset of an OpenAPI 3.0 schema
object, not full JSON Schema. `ChatGoogleAI.for_api/1` passes a `Function`'s
parameters through unmodified, so any keyword outside that subset reaches the
API and is rejected.

`additionalProperties` is the keyword that matters, because
`FunctionParam.to_parameters_schema/1` adds it to every schema it generates.
The result is that **a tool declared the ordinary way, with a `parameters:`
list of `FunctionParam` structs, is rejected by Gemini.** This is not limited
to callers who set the keyword deliberately.

Verified on `main` (cbd3acd) and on v0.9.3.

## Use case, and why this matters

This is the common path, not an edge case. The README's own way of declaring a
tool — `Function.new(%{name: ..., parameters: [FunctionParam.new!(...)]})` —
produces a schema carrying `additionalProperties: false`, and Gemini refuses
the request that carries it. Any tool with at least one parameter is affected.

The rejection is also worse than a per-tool failure. It comes from Google's
protobuf parser, before any semantic validation, so it fails the entire
request. One unsupported keyword in one declaration means no turn runs at all,
and the error names a field path rather than the tool, so the cause is not
obvious from the message.

Callers who write `parameters_schema` by hand hit the same wall from the other
direction. A schema that sets `additionalProperties: false` on purpose — to
refuse unexpected fields — cannot be sent to Gemini at all, even though it is
valid JSON Schema and works on every other provider.

## Context

The keyword arrives from inside the library. `FunctionParam.to_parameters_schema/1`
emits it unconditionally:

```elixir
def to_parameters_schema(params) when is_list(params) do
  %{
    "type" => "object",
    "properties" => Enum.reduce(params, %{}, &to_json_schema(&2, &1)),
    "required" => required_properties(params),
    "additionalProperties" => false
  }
end
```

It recurses, so nested objects and array items carry it too. This was added in
#490 ("fix: add additionalProperties: false to tool schemas") to fix #440,
where Anthropic rejected schemas that lacked it. That change's analysis covered
OpenAI ("not affected") and Anthropic ("all tool/function calling is broken"),
and concluded the fix was "backward compatible ... without affecting other
providers." Google was not evaluated. The Google test was updated in the same
commit to assert the new keyword flows through to Gemini, which is why the
regression has a passing test pinning it.

So the two providers want opposite things on the same field: Anthropic
requires `additionalProperties`, Google refuses it. That is a provider dialect
difference, and the adapter is where the library already resolves those.
`ChatVertexAI` does exactly this today, for the same API family:

```elixir
# Note: We strip the "strict" field as it's OpenAI-specific and not supported by Vertex AI
"functionDeclarations" =>
  functions
  |> Enum.map(&ChatOpenAI.for_api(vertex_ai, &1))
  |> Enum.map(&Map.delete(&1, "strict"))
```

There is also an existing workaround in `ChatGoogleAI` that this bug silently
disabled. Directly below the serialization, a guard deletes `parameters`
entirely when the schema is empty, because Google errors on an empty OBJECT.
It compares against `%{"properties" => %{}, "type" => "object"}` — a shape
without `additionalProperties` — so since #490 it no longer matches what the
generator produces, and functions declared with `parameters: []` stopped
taking that path.

Related: #278 proposes a general schema-translation layer, observing that
providers handle JSON Schema quite differently. It is still open. This issue is
the narrow, concrete instance of it.

## Reproduction

1. Declare a function the ordinary way, with one parameter.
2. Serialize it for Google and look at the emitted declaration, or run any
   chain against a Gemini model with that tool attached.

```elixir
alias LangChain.ChatModels.ChatGoogleAI
alias LangChain.{Function, FunctionParam}

{:ok, weather} =
  Function.new(%{
    name: "get_weather",
    description: "Get the current weather in a given US location",
    parameters: [
      FunctionParam.new!(%{name: "city", type: "string", required: true})
    ],
    function: fn _args, _context -> {:ok, "75 degrees"} end
  })

ChatGoogleAI.for_api(weather)
# => %{
#      "name" => "get_weather",
#      "description" => "Get the current weather in a given US location",
#      "parameters" => %{
#        "type" => "object",
#        "properties" => %{"city" => %{"type" => "string"}},
#        "required" => ["city"],
#        "additionalProperties" => false     # <- rejected by the API
#      }
#    }
```

Sending that declaration returns HTTP 400:

```
Invalid JSON payload received. Unknown name "additionalProperties" at
'tools[0].function_declarations[0].parameters': Cannot find field.
```

Both `true` and `false` are rejected — the parser fails on the unknown field
name, so the value is irrelevant. The same is true for `$schema`, `const`,
`propertyNames`, `patternProperties`, `exclusiveMinimum`, `multipleOf`, and the
other keywords Google's `Schema` type has no field for. The `Schema` type is
identical in `v1` and `v1beta`.

## Observed

Every tool declared with a non-empty `parameters:` list is rejected by Gemini,
and the whole request fails rather than the single tool. A hand-written
`parameters_schema` carrying any keyword outside Google's subset fails the same
way.

## Expected

`ChatGoogleAI` sends a declaration Google accepts, the way `ChatVertexAI`
already strips `strict`. Keywords Google supports — `minimum`, `maximum`,
`enum`, `nullable`, `title`, `pattern`, `default`, `format` — are preserved.

## Offer

I have a fix prepared and can open a PR. It removes the unsupported keywords in
`ChatGoogleAI`'s function-declaration serialization, recursing through
`properties`, `items`, and `anyOf`, keeping everything Google's `Schema` type
does have a field for. It also repairs the empty-parameters guard described
above. Tests cover removal, nested removal, and — to guard against
over-stripping — that supported keywords survive. The existing test asserting
`additionalProperties` reaches Google needs updating, since it pins the
reported behavior.

One consequence has to be stated plainly, because it is a real loss: dropping
`additionalProperties` means Gemini does not enforce closed objects, so a tool
call's arguments may contain fields the schema did not declare. My patch says
so in the moduledoc rather than leaving it silent. If you would rather warn, or
offer an opt-out, I am happy to change it.

There may also be a better fix than stripping, and it is your call rather than
mine. `v1beta` has a `parametersJsonSchema` field on `FunctionDeclaration` that
accepts full JSON Schema, including `additionalProperties`, and is mutually
exclusive with `parameters`. Using it would fix this with no semantic loss at
all. I did not take that path because the field does not exist in `v1`, and
`ChatGoogleAI` lets callers set `api_version` — so it needs a version-conditional
decision about which field to emit, and possibly a fallback, which felt like
yours to make rather than something to slip into a bug fix. If you prefer that
shape, I will redo the PR on it.

lostbean:
Fixes #598.

Google's function declarations accept a select subset of an OpenAPI 3.0 schema
object, not full JSON Schema. A declaration carrying a keyword outside that
subset is rejected by the API:

```
Invalid JSON payload received. Unknown name "additionalProperties" at
'tools[0].function_declarations[0].parameters': Cannot find field.
```

The error comes from the protobuf parser, before semantic validation, so it
fails the whole request rather than the single tool. Both `true` and `false`
are rejected — the parser fails on the unknown field name, so the value does
not matter.

`for_api/1` passed a `Function`'s parameters through unmodified. This affected
far more than schemas that set the keyword deliberately, because
`FunctionParam.to_parameters_schema/1` adds `"additionalProperties" => false`
to every schema it generates and recurses into nested objects and array items.
Any tool declared with the ordinary `parameters:` list form was therefore
rejected by Gemini. That injection came from #490, which fixed #440 for
Anthropic — Anthropic requires the keyword, Google refuses it, and Google was
not evaluated at the time.

This PR removes the unsupported keywords before sending, recursing through
`properties`, `items`, and `anyOf`. The list covers keywords Google's `Schema`
type has no field for. Supported ones — `minimum`, `maximum`, `enum`,
`nullable`, `title`, `pattern`, `default`, `format` — are kept. I used a
denylist rather than an allowlist of the supported fields so that a keyword
Google adds later keeps working instead of being silently dropped; if you would
rather have the stricter allowlist, that is an easy change.

Resolving the difference in the adapter follows what `ChatVertexAI` already
does for the same API family, where it strips `strict` as OpenAI-specific.

This also repairs the empty-parameters guard directly below the change. That
guard deletes `parameters` when the schema is empty, because Google errors on
an empty OBJECT, but it compares against a shape without
`additionalProperties`, so since #490 it had stopped matching what the
generator produces. Sanitizing first makes it fire again for functions declared
with `parameters: []`.

**One removal changes what the API enforces, and I want to be explicit about
it.** Without `additionalProperties`, Gemini does not reject unexpected
properties, so a tool call's arguments may contain fields the schema did not
declare. Callers who relied on that guarantee must validate inside the tool's
own function. The moduledoc states this rather than leaving it silent, since
otherwise the library would quietly claim an enforcement it no longer has. If
you would prefer a warning or an opt-out over a silent strip, say so and I will
add it.

Tests cover three cases: keywords removed from a hand-written
`parameters_schema`, removal from nested schemas, and — to guard against
over-stripping — that supported keywords survive untouched. The existing test
asserting `additionalProperties` reaches Google is updated, since that
assertion pinned the reported behavior.

`ChatVertexAI` reaches the same API through `ChatOpenAI.for_api/2` and has the
same gap. I left it out to keep this to one provider, and can extend the fix if
you want them handled together.

Finally, there may be a better fix than stripping. `v1beta` has a
`parametersJsonSchema` field on `FunctionDeclaration` that accepts full JSON
Schema, including `additionalProperties`, and is mutually exclusive with
`parameters`. That would fix this with no semantic loss. I did not take it
because the field does not exist in `v1` and `ChatGoogleAI` lets callers set
`api_version`, so it needs a version-conditional decision about which field to
emit — a design call that seemed yours rather than something to fold into a bug
fix. Happy to redo this on that shape if preferred.

### DIFF
--- a/lib/chat_models/chat_google_ai.ex
+++ b/lib/chat_models/chat_google_ai.ex
@@ -46,14 +46,16 @@ defmodule LangChain.ChatModels.ChatGoogleAI do
   [select subset](https://ai.google.dev/api/caching#Schema) of an OpenAPI 3.0
   schema object rather than full JSON Schema. Keywords outside that subset are
   rejected by the API with `Unknown name "<keyword>" ... Cannot find field`,
-  which fails the entire request rather than the single tool. This module
-  therefore removes those keywords from a tool's parameters before sending them.
+  which fails the entire request rather than the single tool. Tool parameters
+  and the JSON response schema are therefore passed through
+  `LangChain.Utils.GoogleSchema.sanitize/1` before being sent.
 
   One removal changes what the API enforces: `additionalProperties` is not
   supported, so an object schema that sets `additionalProperties: false` is
   still sent, but without it. Gemini does not reject unexpected properties, and
   a tool call's arguments may contain fields the schema did not declare. Where
-  that matters, validate the arguments inside the tool's own function.
+  that matters, validate the arguments inside the tool's own function. See
+  `LangChain.Utils.GoogleSchema` for the full list of removed keywords.
   """
   use Ecto.Schema
   require Logger
@@ -71,6 +73,7 @@ defmodule LangChain.ChatModels.ChatGoogleAI do
   alias LangChain.TokenUsage
   alias LangChain.LangChainError
   alias LangChain.Utils
+  alias LangChain.Utils.GoogleSchema
   alias LangChain.Callbacks
   alias LangChain.NativeTool
   alias LangChain.Message.Citation
@@ -243,7 +246,7 @@ defmodule LangChain.ChatModels.ChatGoogleAI do
     {response_mime_type, response_schema} =
       case google_ai.json_response do
         true ->
-          {"application/json", google_ai.json_schema}
+          {"application/json", GoogleSchema.sanitize(google_ai.json_schema)}
 
         false ->
           {nil, nil}
@@ -462,13 +465,13 @@ defmodule LangChain.ChatModels.ChatGoogleAI do
     encoded =
       %{
         "name" => function.name,
-        "parameters" => function |> ChatOpenAI.get_parameters() |> sanitize_parameters_schema()
+        "parameters" => function |> ChatOpenAI.get_parameters() |> GoogleSchema.sanitize()
       }
       |> Utils.conditionally_add_to_map("description", function.description)
 
     # For functions with no parameters, Google AI needs the parameters field removing, otherwise it will error
     # with "* GenerateContentRequest.tools[0].function_declarations[0].parameters.properties: should be non-empty for OBJECT type\n"
-    if encoded["parameters"] == %{"properties" => %{}, "type" => "object"} do
+    if GoogleSchema.empty_object?(encoded["parameters"]) do
       Map.delete(encoded, "parameters")
     else
       encoded
@@ -482,49 +485,6 @@ defmodule LangChain.ChatModels.ChatGoogleAI do
   def for_api(%NativeTool{name: name, configuration: nil}) do
     name
   end
-
-  # Keywords that are valid JSON Schema but are not fields of Google's `Schema`
-  # type. A function declaration carrying any of them is rejected outright with
-  # `Unknown name "<keyword>" at 'tools[0].function_declarations[0].parameters':
-  # Cannot find field.`, which fails the whole request rather than the one tool.
-  #
-  # `additionalProperties` is the one that matters most in practice, because
-  # `FunctionParam.to_parameters_schema/1` adds it to every generated schema.
-  @unsupported_schema_keywords ~w(
-    additionalProperties unevaluatedProperties patternProperties propertyNames
-    $schema $ref $defs $comment $id definitions
-    additionalItems prefixItems uniqueItems
-    dependencies dependentRequired dependentSchemas
-    const examples exclusiveMinimum exclusiveMaximum multipleOf
-    readOnly writeOnly deprecated
-    contentEncoding contentMediaType contentSchema
-  )
-
-  # Google's function declarations accept a select subset of an OpenAPI 3.0
-  # schema object, not full JSON Schema. See
-  # https://ai.google.dev/api/caching#Schema for the supported fields.
-  #
-  # Note the semantic loss: `additionalProperties: false` is dropped, so Gemini
-  # does not enforce closed objects and may return unexpected properties in a
-  # tool call's arguments. Validate in the tool itself where that matters.
-  defp sanitize_parameters_schema(%{} = schema) do
-    schema
-    |> Enum.reject(fn {key, _value} ->
-      to_string(key) in @unsupported_schema_keywords
-    end)
-    |> Map.new(fn {key, value} -> {key, sanitize_schema_value(value)} end)
-  end
-
-  defp sanitize_parameters_schema(schema), do: schema
-
-  # Recurse into the containers that hold nested schemas: `properties` maps a
-  # name to a schema, while `items` and `anyOf` hold schemas directly.
-  defp sanitize_schema_value(%{} = value), do: sanitize_parameters_schema(value)
-
-  defp sanitize_schema_value(value) when is_list(value),
-    do: Enum.map(value, &sanitize_schema_value/1)
-
-  defp sanitize_schema_value(value), do: value
 
   @doc """
   Calls the Google AI API passing the ChatGoogleAI struct with configuration, plus
--- a/lib/chat_models/chat_google_ai.ex
+++ b/lib/chat_models/chat_google_ai.ex
@@ -39,6 +39,21 @@ defmodule LangChain.ChatModels.ChatGoogleAI do
   The above call will return the current Google stock price.
 
   When `google_search` is used, the model will also return grounding information in the metadata attribute of the assistant message.
+
+  **Tool Schemas**
+
+  Google's function declarations accept a
+  [select subset](https://ai.google.dev/api/caching#Schema) of an OpenAPI 3.0
+  schema object rather than full JSON Schema. Keywords outside that subset are
+  rejected by the API with `Unknown name "<keyword>" ... Cannot find field`,
+  which fails the entire request rather than the single tool. This module
+  therefore removes those keywords from a tool's parameters before sending them.
+
+  One removal changes what the API enforces: `additionalProperties` is not
+  supported, so an object schema that sets `additionalProperties: false` is
+  still sent, but without it. Gemini does not reject unexpected properties, and
+  a tool call's arguments may contain fields the schema did not declare. Where
+  that matters, validate the arguments inside the tool's own function.
   """
   use Ecto.Schema
   require Logger
@@ -447,7 +462,7 @@ defmodule LangChain.ChatModels.ChatGoogleAI do
     encoded =
       %{
         "name" => function.name,
-        "parameters" => ChatOpenAI.get_parameters(function)
+        "parameters" => function |> ChatOpenAI.get_parameters() |> sanitize_parameters_schema()
       }
       |> Utils.conditionally_add_to_map("description", function.description)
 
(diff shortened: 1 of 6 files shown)


brainlid:
Thanks for this — the diagnosis is right and the writeup made it easy to verify. `FunctionParam.to_parameters_schema/1` does inject `additionalProperties` unconditionally, Google's `Schema` protobuf has no field for it, and the failure is a whole-request 400 rather than a per-tool one. Confirmed on this branch.

I've pushed a commit on top rather than sending you back around. Three things:

**1. `properties` keys were being filtered as if they were schema keywords.**

The recursion treated every nested map as a schema object, but `properties` is a name→schema map whose keys are caller-chosen. A tool with a parameter named `const` or `examples` had that parameter silently deleted:

```elixir
Function.new(%{
  name: "record_config",
  parameters: [
    FunctionParam.new!(%{name: "const", type: :string, required: true}),
    FunctionParam.new!(%{name: "examples", type: :string})
  ],
  ...
})

# before:
# %{"type" => "object", "required" => ["const"], "properties" => %{}}
```

Both parameters gone, and `required` still pointing at one of them. Recursion is now key-aware: `properties` values get sanitized but its keys don't, `items`/`anyOf` hold schemas, and every other value is instance data left alone — otherwise an `enum` member or a `default` value that happens to use a keyword as a field name gets corrupted too.

**2. Extended the denylist.** You asked whether to switch to an allowlist; we kept the denylist for the forward-compat reason you gave, but it was missing `oneOf`, `allOf`, `not`, `if`/`then`/`else`, `contains`/`minContains`/`maxContains`, `unevaluatedItems`, and `$anchor`/`$dynamicRef`/`$dynamicAnchor`/`$vocabulary` — all absent from Google's `Schema`, all producing the same 400. Fixing (1) is what made it safe to add generic words like `not` and `if`.

**3. Took you up on the wider scope.** You flagged `ChatVertexAI` and offered to extend — yes please. Same `Schema` type, same bug. That plus `response_schema` on both models makes four call sites, so the sanitizer moved to `LangChain.Utils.GoogleSchema`. Vertex also never had the empty-parameters guard, so a no-parameter tool still failed with `properties: should be non-empty for OBJECT type`; that's now shared as `GoogleSchema.empty_object?/1` and matches on shape rather than one exact map, so a schema with a stray `description` alongside empty properties is caught too.

On `parametersJsonSchema` — you were right to leave it. It's the better long-term shape, but it's `v1beta`-only and `api_version` is caller-configurable, so it needs a version-conditional decision. Worth a separate issue; this fix should land first since it unblocks the common path on both `v1` and `v1beta`.

On the semantic loss from dropping `additionalProperties`: documenting it is the right call, no warning needed. Gemini wouldn't have enforced it anyway — the request simply failed.

`mix precommit` is green (2115 tests).

brainlid:
Verified against the live Gemini API. Pushed two `live_google_ai` tests covering it.

Worth noting why this got through: the only live tool test we had uses `Calculator`, which declares a `parameters_schema` by hand and never carries `additionalProperties`. Nothing in the suite exercised the `FunctionParam` path — the one the README documents and the one that was broken.

The new tool test names one parameter `const`, so a property whose name collides with a JSON Schema keyword has to survive the round trip. Three runs, three passes.

To confirm the fix is actually what makes the call succeed, I ran the same test against two deliberately broken versions of the sanitizer.

**Sanitizer disabled** — the reported error, exactly:

```
400 INVALID_ARGUMENT
Invalid JSON payload received. Unknown name "additionalProperties" at
'tools[0].function_declarations[0].parameters': Cannot find field.
```

**Sanitizer with the original recursion** (filtering `properties` keys):

```
400 INVALID_ARGUMENT
* GenerateContentRequest.tools[0].function_declarations[0].parameters.required[0]:
  property is not defined
```

That second one is the `properties` bug reaching the wire. The `const` parameter was deleted from `properties` while `required` still named it, so Google rejects the declaration as internally inconsistent. Different error, same dead request — and it's the failure mode the `const` parameter in the test now guards against.

Both tests use `gemini-flash-latest` rather than a pinned version. The alias tracks a current model, which suits a schema-compatibility test where behavior drift doesn't matter but availability does — three other live tests in this file are pinned to `gemini-1.5-flash`, which now 404s, and the module default of `gemini-2.5-pro` has no free-tier quota.

Vertex AI is still unverified live — no `VERTEX_API_KEY`/`VERTEX_API_ENDPOINT` here. The unit tests cover the serialization, and it's the same `Schema` type, but flagging that it's reasoned rather than observed.

`mix precommit` green (2115 tests).

brainlid:
Thanks @lostbean!
❤️💛💙💜

There are still some potential follow-ups in there. Like I'm not setup to verify Vertex AI locally.

PR #600, opened 2026-07-27
Handle raw reasoning-text streaming events in ChatOpenAIResponses

jersearls:
## Problem

Open-weight models that expose their full chain-of-thought rather than a *summary* — notably OpenAI's **`gpt-oss`** family, including when served through AWS Bedrock "Mantle" — stream reasoning as a `reasoning` output item composed of `response.reasoning_text.delta` events, delimited by `response.reasoning_part.added` / `.done` and `response.reasoning_text.done`.

`ChatOpenAIResponses` currently handles the `response.reasoning_summary_*` family but **not** these raw `reasoning_text` / `reasoning_part` events. They fall through to the catch-all in `do_process_response/2` and come back as:

```elixir
{:error, %LangChain.LangChainError{type: "unexpected_response",
  original: %{"type" => "response.reasoning_part.added",
              "part" => %{"type" => "reasoning_text", "text" => ""}, ...}}}
```

for *every* reasoning chunk. That aborts the stream, and because the error tuple is emitted in the delta list, any `on_llm_new_delta` handler that assumes a map crashes with `BadMapError`. Net effect: gpt-oss models are unusable over the Responses API when streaming.

## Fix

- `response.reasoning_text.delta` is now processed exactly like the existing `response.reasoning.delta` handler — emitted as `:thinking` content on a `MessageDelta`, indexed by `output_index` — so callers can display or filter it consistently with other reasoning content.
- The structural markers `response.reasoning_part.added`, `response.reasoning_part.done`, and `response.reasoning_text.done` are skipped, mirroring how `@reasoning_summary_events` are handled.

No breaking changes; purely additive event coverage.

## Testing

- New unit tests: `response.reasoning_text.delta` → `:thinking` `MessageDelta`, and the three structural events → `:skip`. Event shapes match a real gpt-oss-120b stream (captured `reasoning_part.added` with a `reasoning_text` part).
- `mix test test/chat_models/chat_open_ai_responses_test.exs` — 110 tests, 0 failures. `mix format` clean.


brainlid:
Thanks @jersearls! I appreciate the fix!
❤️💛💙💜

PR #608, opened 2026-08-05
Support Responses API context management

fabiokr:
Adds support for the OpenAI Responses API `context_management` option, allowing callers to configure automatic compaction when a conversation reaches a token threshold. This helps long-running conversations stay within the model context window without manually truncating history.

### DIFF
--- a/lib/chat_models/chat_open_ai_responses.ex
+++ b/lib/chat_models/chat_open_ai_responses.ex
@@ -200,12 +200,23 @@ defmodule LangChain.ChatModels.ChatOpenAIResponses do
 
   ## Context Management
 
-  The `context_management` option configures automatic context compaction. For
-  example, compact the context after it reaches 300,000 tokens:
+  The `context_management` option configures server-side context compaction for
+  long-running conversations. When the rendered token count crosses the
+  threshold, the server compacts the context before continuing inference.
+
+  For example, compact the context after it reaches 200,000 tokens:
 
       ChatOpenAIResponses.new!(%{
-        context_management: [%{type: "compaction", compact_threshold: 300_000}]
+        context_management: [%{type: "compaction", compact_threshold: 200_000}]
       })
+
+  `compact_threshold` is optional. When omitted, the server picks a threshold:
+
+      ChatOpenAIResponses.new!(%{context_management: [%{type: "compaction"}]})
+
+  Only the `"compaction"` type is supported today, but entries are passed
+  through to the API as given, so new types and options work without a library
+  change.
 
   ## WebSocket Transport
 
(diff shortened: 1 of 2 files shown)


brainlid:
Thanks for this, @fabiokr. Server-side compaction is a great fit for the long-running agent conversations this library tends to get used for, and the plumbing is right where I'd want it: a field on the schema, key normalization in the changeset, pass-through in `for_api/3`, and registration in `serialize_config/1`. Tests covered the round-trip and the WebSocket envelope, and `mix precommit` was clean on arrival.

I pushed one commit before merging, https://github.com/brainlid/langchain/commit/46571ab. It only touches the validation, which was stricter than the API it wraps. Rather than go off the prose docs, I checked the official SDK type definitions:

```python
# openai-python: src/openai/types/responses/response_create_params.py
class ContextManagement(TypedDict, total=False):
    type: Required[str]
    """The context management entry type. Currently only 'compaction' is supported."""
    compact_threshold: Optional[int]
```

```ts
// openai-node: src/resources/responses/responses.ts
export interface ContextManagement {
  type: string;
  compact_threshold?: number | null;
}
```

That surfaced three cases where we were rejecting requests OpenAI accepts:

1. **`compact_threshold` was required.** It's optional in both SDKs. Omitting it is a valid configuration meaning "let the server pick a threshold", but `[%{type: "compaction"}]` was rejected, and there was a test asserting that rejection.

2. **`type` was pinned to `"compaction"`.** Both SDKs deliberately type it as an open string and put "currently only 'compaction' is supported" in a doc comment instead of an enum. Hard-rejecting anything else means that when OpenAI adds a second type, callers are blocked until we cut a release.

3. **Unrecognized keys were silently dropped.** `normalize_context_management/1` rebuilt each entry from exactly the two known keys, so any new per-entry option disappeared with no error. That's worse than rejecting it, since nothing tells the caller their config didn't take effect.

The rule I try to hold to when wrapping someone else's API: validate against the schema contract, not against the documented usage. Every example in OpenAI's [compaction guide](https://developers.openai.com/api/docs/guides/compaction) passes `compact_threshold`, which I'd guess is why it read as required, but the type says otherwise. Whenever our validation is tighter than the server's, we become the thing blocking a legitimate request, and the caller has no escape hatch.

So the new version normalizes entries to string keys and passes everything else through untouched, then validates only what the API actually constrains: `type` must be a non-empty string, and `compact_threshold`, when present, must be a positive integer.

Tests moved to match. Added accept cases for an entry with no threshold, an entry type we don't know about, and preservation of unknown keys, plus a `for_api/3` case confirming we omit `compact_threshold` rather than sending an explicit null. Kept every rejection case that's genuinely invalid and added a few more: `%{}`, a missing `type`, an atom `type`, and an empty-string `type`.

Also switched the moduledoc example to 200,000 tokens to line up with OpenAI's own example, and documented that the threshold is optional.

`mix precommit` is clean at 2192 tests. Merging shortly. Thanks again for putting this together!

PR #609, opened 2026-08-05
Preserve assistant roles in Responses history

fabiokr:
Preserves assistant roles when serializing conversation history for OpenAI Responses API requests. Previously, assistant messages containing content were rewritten as user messages, changing conversation semantics and potentially causing the model to treat its own prior responses as user instructions.

### DIFF
--- a/lib/chat_models/chat_open_ai_responses.ex
+++ b/lib/chat_models/chat_open_ai_responses.ex
@@ -751,14 +751,26 @@ defmodule LangChain.ChatModels.ChatOpenAIResponses do
         %Message{role: :assistant, content: content} = msg
       )
       when is_list(content) do
+    # Assistant messages are sent back as "output" messages. They only support
+    # `output_text` and `refusal` content parts, so anything else is omitted. If
+    # nothing remains, the message item itself is omitted.
+    assistant_message =
+      case content_parts_for_api(model, content, :assistant) do
+        [] ->
+          []
+
+        parts ->
+          [
+            %{
+              "role" => "assistant",
+              "type" => "message",
+              "content" => parts
+            }
+          ]
+      end
+
     native_tool_calls_for_api(model, content) ++
-      [
-        %{
-          "role" => "assistant",
-          "type" => "message",
-          "content" => content_parts_for_api(model, content)
-        }
-      ] ++
+      assistant_message ++
       Enum.map(msg.tool_calls || [], &for_api(model, &1))
   end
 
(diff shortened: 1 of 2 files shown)


brainlid:
Thanks for catching this. You're right that flattening assistant turns into `"role" => "user"` corrupts the conversation semantics, and you changed exactly the right line.

I've pushed [35efaec](https://github.com/brainlid/langchain/pull/609/commits/35efaeceb24be65a4fb810d834e4e8e94af7aba7) on top of your commit, because the role change alone produces a payload the Responses API rejects.

### Why the extra commit was needed

The Responses API has an asymmetric content-part vocabulary. System and user messages are "input" messages and use `input_text` / `input_image` / `input_file`. Assistant messages are "output" messages and accept only `output_text` and `refusal`.

`content_part_for_api/2` hardcoded `input_text` for every text part regardless of role, so changing just the role produces an item that is internally inconsistent. A live call with any assistant message in the history fails immediately:

```
Invalid value: 'input_text'. Supported values are: 'output_text' and 'refusal'.
```

That would break every Responses API conversation carrying history, so it isn't an edge case. The full unit suite passed on your branch because the tests only assert on the payload we build, never on what OpenAI accepts. That gap is worth noting for anyone touching this module in the future.

This also explains the original `"role" => "user"` rewrite. It was a workaround for the asymmetry rather than an oversight. The old comment recorded the behavior but not the reason, which is what made this look like a one-line fix.

### What the commit does

Content part serialization is now role-aware instead of role-blind:

* `content_parts_for_api/3` and `content_part_for_api/3` take the message role, defaulting to `:user`. Both are public documented functions, so the default keeps existing arity-2 callers working.
* Assistant text serializes to `output_text`.
* Images, files, and other parts have no valid representation inside an assistant message, so they are omitted the same way `:thinking` parts already were.
* If nothing survives that filtering, the message item itself is dropped rather than sending an empty `content` list.

Tests were updated to match: the two new assistant-serialization cases asserted `input_text` and an `input_image` the API won't take, and one pre-existing thinking-parts test needed the same correction. I added coverage for the empty-content path, including the case where an assistant message carries tool calls but no representable content.

`mix precommit` is clean at 2188 passing.

### One caveat

I verified the fix against the live API only up to schema validation. Our OpenAI account ran out of credits partway through, so I could not complete a round trip.

The signal is still meaningful, since it's a controlled comparison. With zero credits, the unfixed payload continues to return the schema validation error, while the fixed payload gets past validation and returns the billing error instead. Validation runs before the credit check, so the corrected payload is schema-valid. I have not confirmed end to end that the model reads the preserved history correctly, and I'll do that before this ships.

PR #619, opened 2026-08-18
Make tool_call_ids unique.

Chuukwudi:
Unlike anthropic's method where they generate the tool_call_id themselves, Google's tool_call_id is implemented here by langchain. Unfortunately it is not unique. the id is just the name of the function.

Sometimes, we might need to switch model providers on one particular conversation.
Anthropic complains if we ask it to continue gemini's conversation because the tool call ids are not unique.

This PR fixes that.

### DIFF
--- a/lib/chat_models/chat_google_ai.ex
+++ b/lib/chat_models/chat_google_ai.ex
@@ -962,11 +962,11 @@ defmodule LangChain.ChatModels.ChatGoogleAI do
 
   def do_process_response(
         _model,
-        %{"functionCall" => %{"args" => raw_args, "name" => name}} = data,
+        %{"functionCall" => %{"args" => raw_args, "name" => name} = call} = data,
         _
       ) do
     %{
-      call_id: Utils.generate_tool_call_id(),
+      call_id: call["id"] || Utils.generate_tool_call_id(),
       name: name,
       arguments: raw_args,
       complete: true,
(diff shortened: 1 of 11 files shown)


brainlid:
Thanks @Chuukwudi, this was a real bug and a good catch. Closing in favor of #620, which carries the fix forward with credit to you.

Two things turned up while digging into it that moved the fix somewhere else.

Google reports an `id` on each `functionCall` part now, confirmed against the live API:

```json
{"functionCall": {"args": {"city": "Denver"}, "id": "call_4162774", "name": "get_weather"}}
{"functionCall": {"args": {"city": "Moab"},   "id": "call_4162775", "name": "get_weather"}}
```

So rather than synthesizing an id, #620 uses the one Google sends and synthesizes only as a fallback, since the field is documented as optional.

The other thing is that unique ids alone did not cover the case you were after. `MessageDelta` pairs tool calls by `index`, and Gemini's parts carry no index, so parallel calls still merged into one during streaming. Gemini turns out to emit each parallel call in its own chunk, which a positional index within a chunk does not fix either. #620 makes a fragment continue the call at its index only when the two ids agree.

Your `Utils` helper idea stayed, as `Utils.generate_tool_call_id/0`, and now backs the synthesized id in five chat models instead of the three separate implementations that were there before.

PR #630, opened 2026-08-21
Carry OpenAI reasoning items across a tool loop

### LINKED ISSUE #618: ChatOpenAIResponses drops reasoning items, so reasoning is not carried across a tool loop
## Summary

The Responses API returns a `reasoning` output item next to the assistant's text
and tool calls. `ChatOpenAIResponses` parses it into a
`%ContentPart{type: :unsupported}`, but nothing ever sends it back, and
`reasoning.encrypted_content` is never captured in the first place. The adapter
defaults to `store: false` and only forces it true when `previous_response_id`
is set, which is the mode OpenAI calls stateless, so the model re-derives its
reasoning on every request, and inside a tool loop that is every round trip.

`ChatAnthropic` replays `thinking` with its `signature` and `redacted_thinking`,
and `ChatGoogleAI` replays `thought_signature`. OpenAI is the one reasoning
provider where continuity silently does not happen.

Verified on `main` (0ce9215, v0.10.0) and on v0.9.6.

## Context

**The inbound parse keeps the id and drops the payload**
(`lib/chat_models/chat_open_ai_responses.ex:2006-2030`):

```elixir
defp content_item_to_content_part_or_tool_call(%{
       "type" => "reasoning",
       "id" => reasoning_id,
       "summary" => summary
     }) do
  # Store reasoning as an unsupported content part for now
  # This preserves the information without breaking the flow
  case ContentPart.new(%{
         type: :unsupported,
         options: %{
           id: reasoning_id,
           summary: summary,
           type: "reasoning"
         }
       }) do
```

In stateless mode, which OpenAI's reasoning guide defines as `store: false` or a
Zero Data Retention organization, the same item also carries
`encrypted_content`, the field that exists so a reasoning item can be handed
back on the next request.

**The outbound path has no clause that can emit it.** `:unsupported` parts are
dropped for every role (`:1016-1021`, under the comment "Ignore unknown,
unsupported content parts"), assistant parts that are not `:text` are dropped
again (`:924`), and the one escape hatch, `native_tool_call_for_api/2`,
recognizes `web_search_call` and nothing else (`:884-891`). A reasoning part
falls through all three and never reaches `input`.

**`ContentPart` documents the opposite behavior.** `lib/message/content_part.ex:16-20`
defines `:unsupported` as a part with no display value because it is encrypted,
"but can be provided back to the LLM to maintain reasoning continuity".
`ChatAnthropic` implements that contract (`chat_anthropic.ex:1989` and `:2003`).
`ChatOpenAIResponses` produces the part the contract describes and then has no
way to fulfil it.

### Where `previous_response_id` helps, and where it does not

It handles the stored case well, and three things fall outside it:

- The adapter sends the full history in `input` alongside the id (reproduction
  (b)), so the stored chain and the resent history both land in context, with no
  way to send only the new items.
- A ZDR organization is in stateless mode by definition ("Stateless mode appli
  when `store` is `false` or when your organization uses Zero Data Retention
  (ZDR)"), so there is nothing stored to chain from.
- Response objects are "saved for 30 days by default", so a long-lived
  conversation resends history anyway.

## Reproduction

Serialization only, no API key needed.

**(a) A reasoning part built exactly as the adapter builds it does not survive.**

```elixir
model = ChatOpenAIResponses.new!(%{model: "gpt-5", api_key: "sk-fake"})

reasoning =
  ContentPart.new!(%{
    type: :unsupported,
    options: %{id: "rs_abc123", summary: [], type: "reasoning"}
  })

msg = Message.new_assistant!(%{content: [reasoning, ContentPart.text!("Let me check.")]})

ChatOpenAIResponses.for_api(model, msg)
#=> [
#=>   %{
#=>     "content" => [
#=>       %{"annotations" => [], "text" => "Let me check.", "type" => "output_text"}
#=>     ],
#=>     "role" => "assistant",
#=>     "type" => "message"
#=>   }
#=> ]
```

The reasoning item is gone, with no warning.

**(b) `previous_response_id` does not stop the history from being sent.**

```elixir
model =
  ChatOpenAIResponses.new!(%{
    model: "gpt-5",
    api_key: "sk-fake",
    previous_response_id: "resp_123"
  })

msgs = [Message.new_user!("first"), Message.new_assistant!("answer"), Message.new_user!("second")]

ChatOpenAIResponses.for_api(model, msgs, []) |> Map.take([:store, :previous_response_id, :input])
#=> %{
#=>   input: [ ...all three messages... ],
#=>   store: true,
#=>   previous_response_id: "resp_123"
#=> }
```

## Expected

- `encrypted_content` is captured into the `:unsupported` part's options when
  present.
- Those parts are serialized back into `input` as reasoning items
  (`%{"type" => "reasoning", "id" => ..., "summary" => [...], "encrypted_content" => ...}`),
  the way `ChatAnthropic.content_part_for_api/1` replays `thinking` and
  `redacted_thinking`.
- A part with no `encrypted_content` is omitted rather than sent as a stub, so
  nothing changes for callers who never had the field.

No new request option is needed. Per OpenAI's reasoning guide, "Reasoning items
in the `output` array will include an `encrypted_content` property containing
encrypted reasoning tokens that you can pass to future calls", and the API
"still accepts the legacy `reasoning.encrypted_content` value in `include` for
compatibility, but doesn't require it". That include value is in any case
already reachable through the existing `:include` field (`:340`).

Replaying the items is the documented stateless path. The guide lists "manually
replay the complete response history" alongside `previous_response_id` as a way
to give a request access to earlier reasoning, and to do that with
`store: false` it says to "preserve every output item, append the next user
message, and replay the complete history".

## Question

Is there a way to get this working today that I have missed? If not, would a PR
be welcome, and which shape would you prefer: `encrypted_content` captured into
the existing `:unsupported` part's options and replayed from there, or somethi
more explicit, such as a dedicated ContentPart type or a field on the model that
turns the replay on

brainlid:
## Problem

The Responses API returns a `reasoning` output item next to the assistant's text and tool calls. `ChatOpenAIResponses` parsed it into a `%ContentPart{type: :unsupported}` and then had no outbound clause that could emit it, and `reasoning.encrypted_content` was never captured in the first place. The model re-derived its reasoning on every request, which inside a tool loop is every round trip.

`ContentPart` documents `:unsupported` as a part that "can be provided back to the LLM to maintain reasoning continuity". `ChatAnthropic` fulfils that contract by replaying `thinking` with its `signature`; `ChatGoogleAI` replays `thought_signature`. This adapter produced the part the contract describes and had no way to send it.

Reported in #618 with a thorough writeup. I verified every claim against `main` before changing anything.

## What the investigation added

The issue covers the non-streaming path. The streaming path is worse: `response.output_item.done` discarded the reasoning `id` too, emitting an empty `:thinking` part. Since agent tool loops usually stream, fixing only the non-streaming path would have left the reported symptom in place for most callers.

That constrained the design. `ContentPart.update_options/2` merges options only when both sides are keyword lists, and `append_content/2` refuses to merge parts of different types. So the streamed reasoning item cannot become an `:unsupported` part at its output index (it would be dropped with a warning), and its options cannot be a map (they would be silently discarded during the delta merge).

## Changes

**Capture** `encrypted_content` on both paths:
- Non-streaming keeps the `:unsupported` part and adds the payload.
- Streaming keeps the `:thinking` part the `output_item.added` event opened at that index, and carries `id`, `summary`, and `encrypted_content` in its options.

**Replay** reasoning items as their own `input` entries. `stand_alone_items_for_api/2` maps assistant content parts to the entries that do not belong inside the message item, preserving order, so a reasoning item precedes the message or function call it produced. `reasoning_item_for_api/1` recognizes both shapes.

**Omit** any reasoning item without `encrypted_content` rather than sending a stub, so nothing changes for callers who never set `include: ["reasoning.encrypted_content"]`.

Reasoning option keys move from a map to a keyword list. That is what `ContentPart` documents, what `merge_part/2` can merge, and it gives the outbound path one shape to read. No test asserted the map shape, and nothing consumed these options before this change, since nothing sent them back.

## Notes for review

- **`:include` is not set automatically.** Without `include: ["reasoning.encrypted_content"]` the API returns no payload and this code stays inert. Auto-adding it for every model risks a 400 on non-reasoning models, so it is documented in the moduledoc instead. Happy to add a narrower auto-enable (only when `:reasoning` options are configured) if you want it.
- **`native_tool_calls_for_api/2` is no longer called internally.** It is public, so I left it rather than make a breaking removal. Say the word and I will drop it.
- **`ContentPart.options` is virtual and excluded from `Jason.Encoder`.** In-memory tool loops are unaffected, but an application persisting messages between processes will lose the payload. Out of scope here; worth knowing.

## Tests

Eleven tests in a new `reasoning item continuity` block. Nine fail on `main`; the other two are regression guards for the omit-without-payload behavior.

Coverage includes capture on both paths, replay of both part shapes, omission without a payload, omission of a plain `:thinking` part, ordering ahead of function calls, ordering preserved when reasoning interleaves with `web_search_call`, a non-streamed parse-then-serialize round trip, and a streamed round trip that merges the deltas into a `Message` and serializes it back.

`mix precommit` is clean: 2316 passed, 0 failures.

Closes #618

### DIFF
--- a/lib/chat_models/chat_open_ai_responses.ex
+++ b/lib/chat_models/chat_open_ai_responses.ex
@@ -111,6 +111,33 @@ defmodule LangChain.ChatModels.ChatOpenAIResponses do
   same usage information for each choice but it is duplicated and only one
   response is meaningful.
 
+  ## Reasoning Continuity
+
+  Reasoning models such as `gpt-5` and the o-series return a `reasoning` output
+  item alongside their text and tool calls. Handing that item back on the next
+  request lets the model resume the reasoning it already did instead of
+  re-deriving it, which matters most inside a tool loop where every tool result
+  is another round trip.
+
+  The item is only replayable when it carries its `encrypted_content`, and the
+  API only returns that when the request asks for it:
+
+      ChatOpenAIResponses.new!(%{
+        model: "gpt-5",
+        include: ["reasoning.encrypted_content"],
+        reasoning: %{effort: :high}
+      })
+
+  With that set, reasoning items are captured off the response and sent back as
+  their own `input` items on subsequent requests. Nothing else is required. A
+  reasoning item that arrived without `encrypted_content` is omitted rather than
+  sent as a stub, so leaving `:include` alone keeps the previous behavior.
+
+  This is the stateless path, and it works whether or not responses are stored.
+  `:previous_response_id` covers the stored case instead, and is unavailable to
+  a Zero Data Retention organization, which the API treats as stateless by
+  definition.
+
   ## Native Tools (Web Search)
 
   Open AI's Responses API also supports built-in tools. Among those, we support Web Search currently.
@@ -809,8 +836,9 @@ defmodule LangChain.ChatModels.ChatOpenAIResponses do
     Enum.map(tool_results, &for_api(model, &1))
   end
 
-  # Native tool calls (such as web_search_call) need to get plucked
-  # out of the content parts and become their own input items.
+  # Reasoning items and native tool calls (such as web_search_call) are their own
+  # entries in the API's `output` array, so they get plucked out of the content
+  # parts and go back as their own input items.
   def for_api(
         %ChatOpenAIResponses{} = model,
         %Message{role: :assistant, content: content} = msg
@@ -834,7 +862,7 @@ defmodule LangChain.ChatModels.ChatOpenAIResponses do
           ]
       end
 
-    native_tool_calls_for_api(model, content) ++
+    stand_alone_items_for_api(model, content) ++
       assistant_message ++
       Enum.map(msg.tool_calls || [], &for_api(model, &1))
   end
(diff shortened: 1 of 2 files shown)


brainlid:
Follow-up: #631 is stacked on this branch. It fixes a separate shape problem in the same file — `web_search_call` and `file_search_call` parts store their options as a map, which does not survive persistence, so they are silently dropped from a restored conversation's history.

Reviewing this one first keeps the diffs clean; #631 will retarget to `main` automatically once this merges.

PR #636, opened 2026-08-26
Stop merging separate streamed responses into one message

brainlid:
## Problem

`LLMChain` turns its accumulated `chain.delta` into a `Message` only when a delta arrives carrying a terminal status. Nothing enforces that a ChatModel ever produces one, and `delta_to_message_when_complete/1` silently no-ops when it does not.

That makes a non-terminating stream fail invisibly. `do_run/1` returns `{:ok, chain}` having appended no message, so three things stay true at once:

1. No assistant message is added.
2. `needs_response` stays `true`, and a looping mode calls the model again.
3. `chain.delta` is cleared only on completion, so it is still live.

The next generation then merges into the same accumulator at the same content index. The user gets one assistant message containing three to five separate, complete answers concatenated end to end, and the count varies with how many calls it took before one finally completed.

`ChatReqLLM` is where this surfaces. `req_llm` derives terminality from a recognized terminal event in the response body (`[DONE]`, `message_stop`, `response.completed`, `response.incomplete`). A stream whose body simply ends carries no such chunk, so every delta stays `:incomplete` and `consume_stream/3` reports success with no completion in it.

This is not specific to one provider. Any adapter that returns a batch of deltas without a terminal one produces the same run-on message.

## Solution

The chain treats an unterminated response as a failed call. The adapter closes a turn itself only where it can prove nothing was lost.

**`LLMChain` never lets a delta outlive the call that produced it.**

`apply_deltas_from_ended_stream/2` wraps the two `do_run/1` delta branches, which were already identical. The streamed response has fully ended by the time `apply_deltas/2` runs there, so a delta still on the chain means no terminal delta ever arrived. That is logged, the delta is dropped, and the call is routed through the existing `handle_delta_error/1` as a new `"incomplete_stream"` error: `:on_llm_error` per attempt, bounded retries from a clean chain, `:on_retries_exceeded` on exhaustion, matching how `delta_conversion_failed` is already handled.

`drop_stale_delta/1` additionally clears any leftover delta at the top of `do_run/1`. A chain can arrive there with one attached, from a caller re-running after an errored stream or from a host that accumulates deltas onto the chain itself, which `usage-rules.md` documents. Dropping it makes cross-call merging structurally impossible rather than merely unreachable.

The partial text is discarded rather than salvaged. The chain cannot tell a response cut off partway through from one that finished and lost only its terminator, and half-streamed text or an incomplete set of tool calls must not enter the transcript.

**`ChatReqLLM` closes a turn only when the finish reason proves it is whole.**

A ChatModel sometimes can tell the difference, and `req_llm` records it: `finish_reason` is `:stop` when a terminal event was seen and `:incomplete` when the body just ended. `consume_stream/3` now checks, once the stream is exhausted, whether any delta carried a terminal status. If none did:

- A reason reporting a finished response (`:stop`, `:tool_calls`, `:length`, `:content_filter`) means only the marker went missing. A closing delta is synthesized with the status that reason maps to and delivered through the streaming callback like any other, so consumers building their own message from `:on_llm_new_delta` see the turn finish too.
- Anything else, including `:incomplete` and a reason that could not be read, leaves the response unfinished for the chain to retry.

Reading the finish reason means calling into a separate `req_llm` process. It is wrapped so a collector that is gone or fails reports no reason, which leaves the turn unfinished rather than closing it on a guess.

Public contracts are unchanged. `apply_deltas/2` and `delta_to_message_when_complete/1` still return incomplete deltas untouched for incremental and mid-stream callers.

## Changes

- [`lib/chains/llm_chain.ex`](https://github.com/brainlid/langchain/blob/me-fix-multiplied-text/lib/chains/llm_chain.ex#L1203) - New `drop_stale_delta/1` and `apply_deltas_from_ended_stream/2`; both `do_run/1` delta branches route through the latter, and `do_run/1` drops a leftover delta before invoking the model.
- [`lib/chat_models/chat_req_llm.ex`](https://github.com/brainlid/langchain/blob/me-fix-multiplied-text/lib/chat_models/chat_req_llm.ex#L448) - `consume_stream/3` appends a closing delta when the stream carried no terminal chunk and the finish reason reports a finished response; new `closing_delta/2`, `close_turn/1`, `stream_finish_reason/1`, and `deliver_deltas/3`.
- [`lib/chat_models/chat_req_llm.ex`](https://github.com/brainlid/langchain/blob/me-fix-multiplied-text/lib/chat_models/chat_req_llm.ex#L53) (`@moduledoc`) - New "Stream Completion" section covering both outcomes.
- [`test/chains/llm_chain_test.exs`](https://github.com/brainlid/langchain/blob/me-fix-multiplied-text/test/chains/llm_chain_test.exs#L2868) - Retry exhaustion (three attempts, callbacks per attempt, error type preserved through exhaustion, `needs_response` still true), a looping mode held to the retry budget under `:while_needs_response` (the mode with no run bound of its own), transient recovery asserting the retried message contains its text exactly once, and the stale-delta drop.
- [`test/chat_models/chat_req_llm_test.exs`](https://github.com/brainlid/langchain/blob/me-fix-multiplied-text/test/chat_models/chat_req_llm_test.exs#L1409) - New describe block covering both branches of the finish-reason decision: closed for `:stop` and `:length`, left unfinished for `:incomplete` and an unreadable reason, callback delivery, no-op when a terminal chunk is present, empty streams, and end-to-end runs through `LLMChain` for both outcomes. Adds a `fake_stream_response/2` helper supplying a real metadata handle.
- [`usage-rules.md`](https://github.com/brainlid/langchain/blob/me-fix-multiplied-text/usage-rules.md#L28) - The streaming flow now records that a response which never terminates is a retried error rather than a silent success.

Two existing tests had encoded the old silent-success behavior in their setup and were updated intent-preservingly. "ignores empty lists in the list of messages" now uses a terminated stream and asserts the merged message, so the fixture represents a real stream and the test still tests what its name says. "remove delta and adds cancelled message" now builds its mid-stream state through the public `merge_delta/2`, matching how a real cancellation host accumulates deltas, since cancelling happens partway through a stream rather than after a completed run.

## Call volume

The old path multiplied calls *because* it concatenated. The hanging delta is what kept `needs_response` true, and that is what re-entered the mode loop, so the two were the same defect seen from either end.

| | Before | After |
|---|---|---|
| What drives the re-call | `needs_response` still true, so the mode loop re-enters | `handle_delta_error/1`, the chain's existing failure path |
| Bound | None. `WhileNeedsResponse` has no `check_max_runs` | `max_retry_count`, default 3, per turn |
| Delta between calls | Left live, so call N+1 merges into call N's accumulator | Dropped before every attempt |
| Result | One message, N generations concatenated | One clean message, or one error |
| Observability | None. Reported as success | `:on_llm_error` per attempt, `:on_retries_exceeded` on exhaustion |

A response that finished and lost only its terminal marker is now one call that succeeds, since the adapter closes the turn from the finish reason and never reaches the retry path. Only a stream reported as genuinely cut short retries, and that response really is a fragment.

## Testing

`mix precommit` is clean: 2372 tests pass (41 doctests), no formatting, Credo, or dependency-audit issues. `mix dialyzer` passes. No live API calls are involved; the new tests stub `ReqLLM.stream_text/3` with Mimic and drive `LLMChain` with a stubbed `ChatOpenAI.call/3`, following the existing conventions in both files.

## Notes for reviewers

- A retry replays. The partial text reached consumers through `:on_llm_new_delta` before the retry started, so a host that appends deltas to a live buffer shows both attempts until the final message replaces it. `ChatReqLLM.handle_request_failure/6` declines to retry for exactly this reason at the transport layer. Retrying here is still the right call, since the alternative is admitting a fragment into the transcript, but hosts that render raw deltas should reset their buffer on `:on_llm_error`.
- `LangChain.Chains.LLMChain.Modes.WhileNeedsResponse` has no run bound, unlike `UntilToolUsed`, whose pipeline includes `check_max_runs/2`. `UntilSuccess` is similarly exposed, since it recurses on any last message that is not an assistant response or a clean tool result. Both are made safe against this failure by the change above, but adding a default `max_runs` would break legitimate long tool loops and deserves its own decision.
- `ChatReqLLM` assigns one content index per chunk type rather than per content block, so several text blocks in a single response merge into one. That cannot be fixed here yet: `req_llm`'s Responses decoder threads an output index through tool-call chunks but not text ones, so separate blocks are indistinguishable downstream.

### DIFF
--- a/lib/chains/llm_chain.ex
+++ b/lib/chains/llm_chain.ex
@@ -297,9 +297,9 @@ defmodule LangChain.Chains.LLMChain do
     # Track the current merged `%MessageDelta{}` struct received when streamed.
     # Set to `nil` when there is no current delta being tracked. This happens
     # when the final delta is received that completes the message. At that point,
-    # the delta is converted to a message and the delta is set to nil. A response
-    # that ends without a completing delta is converted anyway, so a delta never
-    # outlives the LLM call that produced it.
+    # the delta is converted to a message and the delta is set to nil. An LLM
+    # call neither starts nor finishes with a delta attached, so a delta never
+    # outlives the call that produced it.
     field :delta, :any, virtual: true
 
     # Track the last `%Message{}` received in the chain.
@@ -924,6 +924,8 @@ defmodule LangChain.Chains.LLMChain do
   end
 
   defp do_run(%LLMChain{} = chain) do
+    chain = drop_stale_delta(chain)
+
     # submit to LLM. The "llm" is a struct. Match to get the name of the module
     # then execute the `.call` function on that module.
     %module{} = chain.llm
@@ -961,11 +963,11 @@ defmodule LangChain.Chains.LLMChain do
 
       {:ok, [%MessageDelta{} | _] = deltas} ->
         if chain.verbose_deltas, do: IO.inspect(deltas, label: "DELTA MESSAGE LIST RESPONSE")
-        apply_llm_deltas(chain, deltas)
+        apply_deltas_from_ended_stream(chain, deltas)
 
       {:ok, [[%MessageDelta{} | _] | _] = deltas} ->
         if chain.verbose_deltas, do: IO.inspect(deltas, label: "DELTA MESSAGE LIST RESPONSE")
-        apply_llm_deltas(chain, deltas)
+        apply_deltas_from_ended_stream(chain, deltas)
 
       {:error, %LangChainError{} = reason} ->
         if chain.verbose, do: IO.inspect(reason, label: "ERROR")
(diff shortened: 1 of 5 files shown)


brainlid:
This is a merge of https://github.com/brainlid/langchain/pull/637 by @edraut as well.

PR #637, opened 2026-08-26
fix handle failed streams and drop stale deltas

edraut:
## Problem

When a streamed LLM response ends without a terminal chunk (the provider stream
was cut off, or ended with a frame the provider decoder doesn't recognize —
e.g. the OpenAI Responses API's `response.failed`), the merged delta never
reaches `status: :complete`. `apply_deltas/2` returns `{:ok, chain}` with no
message added and the hanging delta still on the chain, so `do_run/1` reports
the LLM call as a **success that produced nothing**.

Two bad things follow:

1. `needs_response` never updates (it only changes when a message is added), so
   agent loops built on it re-invoke the LLM with the identical transcript —
   in our production case, up to `max_runs` times.
2. The hanging delta is never cleared, so each re-invocation's text merges onto
   the previous attempt's — the eventual message contains the answer repeated
   N times (or fails `to_message/1` with `delta_conversion_failed`).

## Fix

- `do_run/1`'s streaming arms now check the chain after `apply_deltas/2`: a
  delta still present means the stream ended without a terminal. That case is
  logged and routed through the existing `handle_delta_error/1` as a new
  `"incomplete_stream"` error type — so it gets `:on_llm_error`, bounded
  retries from a clean chain (`max_retry_count`), and `:on_retries_exceeded`,
  matching how `delta_conversion_failed` is handled.
- `do_run/1` also drops (with a warning) any leftover delta from a prior
  errored call before invoking the model, so cross-call delta merging is
  structurally impossible.

Public contracts are unchanged: `apply_deltas/2` and
`delta_to_message_when_complete/1` still return incomplete deltas untouched
for incremental/mid-stream callers. The partial text is deliberately discarded
on retry rather than salvaged — half-streamed text or tool-call JSON must not
enter the transcript.

## Tests

Three new tests in `llm_chain_test.exs`: retry exhaustion (3 attempts,
callbacks fired per attempt, error type preserved), transient recovery
asserting the retried message contains its text exactly once (the duplication
regression), and the stale-delta drop.

Two existing tests had encoded the old silent-success behavior in their setup
and were updated intent-preservingly: "ignores empty lists in the list of
messages" now uses a terminated stream and asserts the merged message;
"remove delta and adds cancelled message" now builds its mid-stream state via
the public `merge_delta/2` (matching how a real cancellation host accumulates
deltas) instead of running an unterminated stream through `run/1`.
```

### DIFF
--- a/lib/chains/llm_chain.ex
+++ b/lib/chains/llm_chain.ex
@@ -922,6 +922,8 @@ defmodule LangChain.Chains.LLMChain do
   end
 
   defp do_run(%LLMChain{} = chain) do
+    chain = drop_stale_delta(chain)
+
     # submit to LLM. The "llm" is a struct. Match to get the name of the module
     # then execute the `.call` function on that module.
     %module{} = chain.llm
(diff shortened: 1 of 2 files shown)


brainlid:
Thanks @edraut! I combined your approach in with PR #636 and merged it.
❤️💛💙💜

PR #644, opened 2026-09-04
Document using ChatOllamaAI with llmman

ericcurtin:
[llmman](https://github.com/llmmanorg/llmman) is a local model runner that serves the Ollama API on port 17434.

- No new chat model module: llmman's `/api/chat` is the Ollama chat API, so `ChatOllamaAI` works unchanged with `endpoint: "http://localhost:17434/api/chat"`. This mirrors the existing "Cloudflare Workers AI via `ChatOpenAI`" pattern instead of duplicating the module.
- README: one bullet in the supported chat models list and a short `### llmman` section showing the endpoint override.
- `ChatOllamaAI` moduledoc: a `### Using with llmman` subsection so hexdocs covers it too.
- No `provider/0`, telemetry, CHANGELOG or test changes since no code path changes; happy to add a CHANGELOG line if preferred.

**Testing:** `mix format --check-formatted` and `mix test test/chat_models/chat_ollama_ai_test.exs` (48 passed) pass; docs-only change.

> AI-assisted, reviewed before submitting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,7 @@ Elixir LangChain enables Elixir applications to integrate AI services and self-h
 - **Google Vertex AI** - Google's enterprise AI offering
 - **DeepSeek** - DeepSeek models with prompt caching support
 - **Ollama** - Locally hosted open-source models
+- **llmman** - Local model runner serving the Ollama API on port 17434, via `ChatOllamaAI`
 - **Mistral** - Mistral AI models
 - **Perplexity** - Perplexity AI models
 - **orq.ai** - orq.ai Deployments API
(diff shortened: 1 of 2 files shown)


brainlid:
Thank you for this, and for checking it against `ChatOllamaAI` before sending it over. The write-up was clear and the endpoint-override observation is correct.

I'm going to pass on it, though, and the reason is scope rather than quality. There is a growing set of local runners and compatible gateways out there, and each one that lands in the README becomes something I have to keep accurate as its ports, env vars and model names move around. I don't have a good way to stay current with all of them, and documentation that has quietly gone stale is worse for people than no documentation at all. So I'd rather the README teach the general pattern and leave the per-runner specifics to each runner's own docs, where they will actually be maintained.

The good news is that the general pattern got quite a bit stronger this week. https://github.com/brainlid/langchain/pull/648 just landed, which broadens `ChatOpenAI` for OpenAI-compatible interfaces. The main addition is `extra_body`, a map merged into the request as the final step, so you can send parameters this library has no struct field for, and pass `nil` to remove one the model always sends that your server rejects. It also adds `service_tier` pass-through and correct `max_completion_tokens` handling.

Since llmman serves OpenAI on the same 17434 endpoint as Ollama, `ChatOpenAI` is probably the better door in for llmman users regardless. It has an `api_key` field, which `ChatOllamaAI` does not, so it also covers the case where the daemon is reachable on the network and expects a key on every request.

Thanks again for taking the time to put this together. No reflection on llmman itself.

PR #647, opened 2026-09-09
Support OpenAI Bedrock Chat Completions options

akravetz:
## Problem

`ChatAwsMantle` rejects `reasoning_effort: "none"`. It accepts only `low`,
`medium`, and `high`.

Some OpenAI models require `"none"` when a Chat Completions request contains
function tools. The adapter can add this value to the request. However,
`ChatAwsMantle.new/1` rejects the value first.

`ChatAwsMantle` also sends `max_tokens` by default. Current OpenAI models on
Amazon Bedrock reject this field. They accept `max_completion_tokens`.

AWS provides the following documentation:

- [OpenAI model request parameters](https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-openai.html)
  lists `max_completion_tokens` and maps it to the Converse `maxTokens` field.
- [GPT-5.6 Luna](https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-openai-gpt-56-luna.html)
  lists the model IDs and URLs for both Bedrock endpoints.
- [Bedrock endpoints](https://docs.aws.amazon.com/bedrock/latest/userguide/endpoints.html)
  explains the differences between `bedrock-runtime` and `bedrock-mantle`.

## Change

- Accept `none` as a reasoning effort value.
- Add the optional `max_completion_tokens` field.
- Omit `max_tokens` when `max_completion_tokens` has a value.
- Keep the current `max_tokens: 4096` default for all other requests.
- Add module documentation and unit tests.

This change does not change existing requests unless a caller sets
`max_completion_tokens`.

## Investigation

We sent small synthetic requests to GPT-5.6 Luna through both Bedrock endpoint
families. We used the model ID that AWS documents for each endpoint.

| Endpoint | Model | Token field | Result |
| --- | --- | --- | --- |
| `bedrock-runtime` | `us.openai.gpt-5.6-luna` | `max_tokens` | HTTP 400: unsupported parameter |
| `bedrock-runtime` | `us.openai.gpt-5.6-luna` | `max_completion_tokens` | HTTP 200 |
| `bedrock-mantle` | `openai.gpt-5.6-luna` | `max_tokens` | HTTP 400: unsupported parameter |
| `bedrock-mantle` | `openai.gpt-5.6-luna` | `max_completion_tokens` | HTTP 200 |

We also tested client-side function tools through `bedrock-runtime`. The API
returned HTTP 400 when the request did not disable reasoning. The same tool
request succeeded with `reasoning_effort: "none"`. GPT-5.6 Luna, Sol, and Terra
each called the test tool and returned the correct result.

## Verification

- `mix precommit`: 2,498 tests and 41 doctests passed. The suite excluded 152
  live tests.
- `mix dialyzer --format dialyxir`: passed.
- The live request results are listed above.

### DIFF
--- a/lib/chat_models/chat_aws_mantle.ex
+++ b/lib/chat_models/chat_aws_mantle.ex
@@ -101,8 +101,47 @@ defmodule LangChain.ChatModels.ChatAwsMantle do
     reasonable starting defense.**
   - `:presence_penalty` — -2.0 to 2.0. Binary variant of frequency_penalty
     (penalizes any token that has appeared at all)
-  - `:max_completion_tokens` — Some OpenAI models use this field instead of
-    `:max_tokens`. This field replaces `:max_tokens` in the request.
+
+  ## Token Limits
+
+  `:max_tokens` sets the upper bound on generated tokens. It is sent as
+  `max_completion_tokens`, the name AWS documents for Mantle. The Kimi and
+  gpt-oss models accept either name; OpenAI's newer models reject the older
+  `max_tokens` and accept only this one.
+
+      ChatAwsMantle.new!(%{model: "openai.gpt-oss-120b", max_tokens: 4096})
+
+  The default is `4096`, low enough to cap a runaway generation. Raise it when
+  a reasoning budget needs more headroom.
+
+  A model that accepts only the older `max_tokens` key needs the key swapped
+  with `:extra_body`:
+
+      ChatAwsMantle.new!(%{
+        model: "some.legacy-model",
+        max_tokens: 4096,
+        extra_body: %{"max_completion_tokens" => nil, "max_tokens" => 4096}
+      })
+
+  Keep the `:max_tokens` field set as well, since telemetry reports the limit
+  from it.
+
+  ## Provider-Specific Parameters
+
+  Mantle hosts models from several vendors, each of which accepts parameters
+  this module has no field for. `:extra_body` is a map of values merged into
+  the request body last, so they override anything the model computed:
+
+      ChatAwsMantle.new!(%{
+        model: "moonshotai.kimi-k2.5",
+        extra_body: %{"repetition_penalty" => 1.05}
+      })
+
+  Keys may be strings or atoms, a `nil` value removes the key from the body,
+  and nested maps merge key by key. See `LangChain.Utils.merge_extra_body/2`
+  for the full rules. Overriding `stream`, `messages`, `tools` or `model` can
+  break response handling; headers and transport options belong in
+  `:req_config`, not `:extra_body`.
 
   ## Streaming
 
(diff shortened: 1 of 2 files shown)


brainlid:
Thanks for this, and especially for the investigation table. Both problems you found are real. I pushed an adaptation on top of your commit and rebased onto `main`, because [#648](https://github.com/brainlid/langchain/pull/648) landed in the meantime and set a direction for exactly this question.

**Your `reasoning_effort: "none"` change is kept as-is.** The whitelist was rejecting the value before a request could ever reach AWS, and that was a straightforward bug.

**The token limit is handled differently now.** #648 made `ChatOpenAI` treat `max_completion_tokens` as a *serialization detail* rather than a second option: `:max_tokens` stays the single public field, `for_api/3` writes it out under the current name, and `:extra_body` is the escape hatch for a server that still wants the old key. Carrying a separate `:max_completion_tokens` field on `ChatAwsMantle` would give the library two different mental models for the same concern.

There was also a concrete problem with the two-field version. Telemetry reads `:max_tokens` to report `gen_ai.request.max_tokens`:

https://github.com/brainlid/langchain/blob/main/lib/chat_models/chat_model.ex#L294

So a caller who set only `:max_completion_tokens` sent one limit on the wire while OpenTelemetry reported the untouched `4096` default.

**What the branch does now:**

- `:max_tokens` is the only field, always serialized as `max_completion_tokens`.
- `:extra_body` is added, wired through `LangChain.Utils.merge_extra_body/2`, same as `ChatOpenAI` and `ChatOpenAIResponses`. It covers the reverse swap for a model that only accepts the legacy key, and the vendor-specific parameters that come with Mantle hosting several model families.
- Module docs gain "Token Limits" and "Provider-Specific Parameters" sections mirroring `ChatOpenAI`.

**On the risk of renaming unconditionally.** Mantle is multi-vendor, so the thing your table did not cover is whether the *existing* models still work under the new key. I checked that against Bedrock before making the change:

| Model | `max_completion_tokens` | `max_tokens` |
| --- | --- | --- |
| `moonshotai.kimi-k2.5` | HTTP 200 | HTTP 200 |
| `moonshotai.kimi-k2-thinking` | HTTP 200 | HTTP 200 |
| `openai.gpt-oss-120b` | HTTP 200 | HTTP 200 |

All of them accept both names, so the rename is safe and the `max_tokens: 4096` guard against Kimi's repetition loops still applies.

**One thing I could not verify.** My account does not have GPT-5.6 access, so `openai.gpt-5.6-luna` and `openai.gpt-5.6-sol` both return `401 access_denied` for me rather than a parameter error. Your finding that they reject `max_tokens` matches AWS's published parameter documentation, so I am taking it as given. If you still have access, could you confirm the branch works against Luna as it stands now? That is the one cell in the matrix resting on your testing rather than mine.

`mix precommit` passes at 2426 tests, and the live Mantle suite passes at 49 including streaming, multimodal, and tool calling.

akravetz:
makes total sense! I'll get this tested for ya right now to confirm

akravetz:
confirmed - working, good to go

brainlid:
> confirmed - working, good to go

Thanks @akravetz!

brainlid:
Thanks @akravetz!
❤️💛💙💜

akravetz:
happy to help! good call on the telemetry. I'll keep an eye on that moving forward. I'm starting to flex a couple mantle based llm's so may have some more PR's coming up :)

brainlid:
Nice! Also, I just published this in v0.14.0.

PR #652, opened 2026-09-14
fix: Serialize tool results in OpenTelemetry messages

georgeguimaraes:
`Message.tool_results` wasn't included in OpenTelemetry message serialization, so a completed tool call appeared in `gen_ai.input.messages` as `{"role":"tool","content":null}`. That loses the result content and the call ID needed to connect it to the assistant's tool request.

Expand each tool result into its own message with `role`, `tool_call_id`, and LLM-visible `content`, preserving conversation order. This follows the existing message shape and handles multiple results, text and multimodal content, including error messages. Application-only result fields stay excluded.

Tool results are captured as part of LLM history only when `capture_input_messages` is enabled. Tool-span capture options are unchanged. Thinking and unsupported content parts still use the existing exclusion rules. Reasoning capture would need a separate change with an explicit option.

All three new regression tests failed before the fix. The OpenTelemetry suite now passes all 185 tests. The full suite passed 2,426 of 2,429 tests, with three unrelated failures because `/etc/hostname` isn't available on macOS.


brainlid:
Thanks @georgeguimaraes!
❤️💛💙💜

PR #654, opened 2026-09-14
fix: Make prompt template security tests portable

georgeguimaraes:
Three prompt-template security tests fail on macOS because they read `/etc/hostname`. Use per-test ExUnit temporary files with known contents for the file-read payloads, and let ExUnit manage scratch-file cleanup.

The tests still prove that trusted template code can read a file, while injected code in assigns and composed templates stays literal. The trusted read now checks exact contents instead of just a non-empty result.


brainlid:
Thanks @georgeguimaraes! Nice one!
❤️💛💙💜

PR #655, opened 2026-09-16
fix: Remove unreachable message serialization clauses

georgeguimaraes:
removes five Elixir 1.20 warnings without changing message serialization.

Strict compilation and 226 model tests pass, while the full suite has three existing macOS failures already covered by #654.


brainlid:
Thanks @georgeguimaraes! After seeing what you started with here, I went ahead and upgraded to use the latest Elixir and did the cleanup in this PR https://github.com/brainlid/langchain/pull/658.

